PR #7, opened 2020-11-30
test(visual-regressions): bump reg-cli and remove workaround

abykovssc:
reg-cli used to have a bug which caused it to not update visual regressions correctly, so used a script as workaround. Issue was fixed since then, so workaround can be safely removed


ajkl2533 (approved):
LGTM

abykovssc:
Well, this PR somehow broken itself 🤦  the commits from it are merged into alpha branch, but it says there conflicts 🤦  but doesn't show what kind of conflicts

PR #10, opened 2020-12-04
fix: unify border styles for Card and Collapsible

ajkl2533:
Closes FEP-1461


abykovssc (approved):

abykovssc:
LGTM, but what's surprising is that --disableCssAnimation actually worked for the for loading card in visual regressions 🤔  Which means we can probably now auto test our whole storybook in ui-core 🤔

PR #11, opened 2020-12-07
feat(button): add type prop

vaclav-zeman:
Just to fix this:
Screen Shot 2020-12-07 at 09 57 34

### DIFF
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -28,7 +28,9 @@ const spinnerSizes = {
   [ButtonSizes.sm]: 10,
 };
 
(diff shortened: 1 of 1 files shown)


ajkl2533:
I think you can do just 
```js
const Button: React.FC<
  Omit<React.HTMLProps<HTMLButtonElement>, 'size'> & ButtonProps
> = ({...
```
and that should do the trick. But we have to omit the `size` prop from HTML because it clashing with our own property

vaclav-zeman:
@ajkl2533 nice, changed

ajkl2533 (approved):
LGTM

PR #15, opened 2020-12-09
feat(Hexgrade): add margin prop

vaclav-zeman:


src/components/HexGrade/HexGrade.tsx
```diff
@@ -2,10 +2,13 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { defaultTo, path, pipe } from 'ramda';
 import { isNotUndefined } from 'ramda-adjunct';
+import styled from 'styled-components';
 
 import colors from '../../theme/colors';
 import { HexGradeGrades, HexGradeVariants } from './HexGrade.enums';
 import { HexGradeProps } from './HexGrade.types';
+import { SpacingSizeValuePropType } from '../../types/spacing.types';
```
ajkl2533, line 10: could you please move these two imports before the siblings imports?

src/components/HexGrade/HexGrade.tsx
```diff
@@ -65,8 +73,9 @@ const HexGrade: React.FC<HexGradeProps> = ({
       : colors.graphite5H;
 
   return (
-    <svg
+    <StyledSVG
       height={size}
+      margin={margin}
```
ajkl2533, line 78: I'm just worried that here the `auto` value will be missed a lot :/

src/components/HexGrade/HexGrade.stories.tsx
```diff
@@ -33,6 +33,9 @@ playground.argTypes = {
   isInversed: {
     control: 'boolean',
   },
+  margin: {
+    control: 'object',
```
ajkl2533, line 37: I usually disable controls for `margin` since it's not very user-friendly to edit this in Storybook

ajkl2533 (approved):
Small nitpicks, but LGTM

src/components/HexGrade/HexGrade.tsx
vaclav-zeman: Eventually yeah but FlexContainer is usually good enough for now

src/components/HexGrade/HexGrade.stories.tsx
vaclav-zeman: Agree it's not ideal but no one forces users to edit that - I think it's useful at least for me!

PR #16, opened 2020-12-10
feat(Link): add secondary link color

anakreon:
Ref FEP-1446


src/components/typography/Link/Link.tsx
```diff
@@ -15,7 +15,7 @@ const StyledLink = styled.a.attrs(() => ({
   border: none;
   font-family: inherit;
   font-weight: normal;
-  text-decoration: none;
+  text-decoration: ${getButtonColor('decoration')};
```
ajkl2533, line 18: I don't think retrieving text-decoration value with `getButtonColor` is the correct approach.  Also text-decoration in context of buttons doesn't make much sense to me. I personally would move text-underline in the typography part of theme.

src/theme/buttons.ts
```diff
@@ -30,11 +30,19 @@ const buttons: Buttons = {
...
       activeColor: colors.dietBlueberry,
+      decoration: 'none',
     },
     danger: {
       color: colors.cherry,
       hoverColor: colors.strawberry,
       activeColor: colors.pumpkin,
+      decoration: 'none',
+    },
+    footnote: {
```
ajkl2533, line 41: I would use `secondary` instead of `footnote` because it's a more generic name and we were used to having secondary in this grey-ish colours

ajkl2533 (changes requested):

src/components/typography/Link/Link.tsx
anakreon: yeah, I was thinking whether we'll also want to have 'footprint' text button? since Link styles copy button styles, maybe we want to have 'footprint' text button... if not, I can move this elsewhere...

src/theme/buttons.ts
anakreon: I'd rather use semantic naming, 'secondary' doesn't really say much 🤷. But unless there's other opinions, i suppose you're the boss 😄

src/components/typography/Link/Link.tsx
anakreon: so in fact, it may be better to completely separate button & link styles for clarity and consistency...

ajkl2533: I don't think we need "footnote" button just yet so I don't have any problem if we move the colours from Buttons into Typography, but I'd like to have that connection between those styles (so reference primary colours from buttons in typography). Then we can introduce method `getLinkStyle` (similar to `getFormStyle`) to retrieve the theme values.

src/theme/buttons.ts
ajkl2533: well, 'footnote' has some semantic meaning, but it doesn't have to be true for other applications and also what happens when you'll need to use this style in the header or somewhere else than footnote :D

anakreon: Yeah, that's actually very fair point. And I see that 'secondary' is also used for the text element of the same color (not footnote), so 'secondary' it is 👍

src/theme/typography.ts
```diff
@@ -72,12 +74,34 @@ const margin: Margin = {
   },
 };
 
+const links: Links = {
+  primary: {
+    color: colors.blueberryClassic,
```
anakreon, line 79: even though `primary` and `danger` use the same colors, i didn't want to reuse `button` colors since it would increase coupling between the files. imports between files of the same level of file hierarchy are dangerous and can create spaghetti code (unless there's some reasonable underlying semantics - `typography -> import from 'buttons'` = NONO, but `typography -> import from 'color-base' and buttons -> mport from 'color-base'` could work, if we really want to 'deduplicate' this).
what do you think @ajkl2533?

ajkl2533: OK leave it this way for now 👍

src/theme/typography.ts
```diff
@@ -72,12 +74,34 @@ const margin: Margin = {
...
+  },
+  danger: {
+    color: colors.cherry,
+    hoverColor: colors.strawberry,
+    activeColor: colors.pumpkin,
+    decoration: 'none',
+  },
+  secondary: {
+    color: colors.graphite2B,
+    hoverColor: colors.radiantBlueberry,
```
ajkl2533, line 92: Did you check hover and active colors with @MateCziner? It doesn't seem right that we use blue shades for this

anakreon: Yep, he was so nice as to define all the proper styles. Check here:
https://www.figma.com/file/vnzYfrwIISRYkr2iYE5TcD/Platform-%3A-Atoms?node-id=1972%3A491

ajkl2533 (approved):

ajkl2533:
@anakreon could you please add `getLinkStyle` docs in here https://github.com/securityscorecard/design-system/blob/main/src/utils/helpers.stories.mdx

anakreon:
> @anakreon could you please add `getLinkStyle` docs in here https://github.com/securityscorecard/design-system/blob/main/src/utils/helpers.stories.mdx

Yeah, I'm thinking if it makes sense to include these two in the 'utils/helper' file... They seem quite specific to link / button 🤷‍♂️. But i suppose we can clean that later if it's cluttered

PR #18, opened 2020-12-10
fix(router): make react-router-dom optional

ajkl2533:
This PR is making `react-router-dom` optional and dynamic dependency, that means you don't have to have router included in your project e.g. Next is using its own router and make `react-router-dom` unnecessary.
When `react-router-dom` is not installed and you try to use either Button or Link component with `to` property and without specified component or tag in `as` property library will throw an error in the console with possible solutions and will not render anything in DOM.

There is one caveat with this solution if you're using this library without `react-router-dom` it will show a warning about missing module in the console. I haven't found a solution to this but a user can disable the warning in Webpack in his application.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,9 +31,8 @@
     "chromatic": "npx chromatic storybook:build"
   },
   "peerDependencies": {
-    "prop-types": "^15.7.2",
-    "react": "^16.8.4",
-    "react-dom": "^16.8.4",
+    "react": "^16.8.0",
+    "react-dom": "^16.8.0",
     "react-router-dom": "^5.2.0",
     "styled-components": "^5.1.0"
   },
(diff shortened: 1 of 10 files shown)


lhorak (approved):
Looks good!

package.json
```diff
@@ -31,9 +31,8 @@
     "chromatic": "npx chromatic storybook:build"
   },
   "peerDependencies": {
-    "prop-types": "^15.7.2",
-    "react": "^16.8.4",
-    "react-dom": "^16.8.4",
+    "react": "^16.8.0",
+    "react-dom": "^16.8.0",
     "react-router-dom": "^5.2.0",
```
lhorak, line 36: should the reaxt-router-dom stil be here?

ajkl2533: Yep it's still peer dependency, but now you don't need to have it installed

lhorak: ok

youssefssc (approved):

PR #22, opened 2020-12-14
feat(FullscreenModal): add component

ajkl2533:
This PR is adding a component for FullscreenModal pattern. 

#### Supported layouts

- `single-6` - single column layout (total width of 6 columns)
- `single-8` - single column layout (total width of 8 columns)
- `sidebar-4-6` - two columns layout (total width of 10 columns, 4 for sidebar and 6 for content)
- `sidebar-4-8` - two columns layout (total width of 12 columns, 4 for sidebar and 8 for content)

#### Usage
```jsx
<FullscreenModal
  layout=LAYOUT_TYPE
  header=REACT.NODE // string or component
  content=REACT.NODE // string or component
  footer=REACT.NODE // string or component
  sidebar=REACT.NODE // string or component
/>
```


src/components/FullscreenModal/Footer/useStickyFooter.ts
```diff
@@ -0,0 +1,75 @@
...
+    if (modalRef.current === null) return;
+    const isScrollable =
+      modalRef.current.scrollHeight > modalRef.current.offsetHeight;
+    setShouldShowScrollToTopButton(isScrollable);
+    setIsFixed(false);
+  }, [modalRef, setShouldShowScrollToTopButton]);
+
+  useEffect(() => {
+    showScrollToTopButton();
+    window.addEventListener('resize', debounce(showScrollToTopButton, 200));
```
abykovssc, line 23: we also should remove listener when hook caller is unmounted

src/components/FullscreenModal/Footer/useStickyFooter.ts
```diff
@@ -0,0 +1,75 @@
...
+        subtree: true,
+        attributes: true,
+      });
+
+      isInView();
+      modalRef.current.addEventListener('scroll', debounce(isInView, 100));
+      window.addEventListener('resize', debounce(isInView, 200));
+    }
+    return () => {
+      observer.disconnect();
```
abykovssc, line 61: `scroll` and `resize` listeners also have to be removed here

src/components/FullscreenModal/Footer/useStickyFooter.ts
```diff
@@ -0,0 +1,75 @@
...
+    shouldShowScrollToTopButton,
+    setShouldShowScrollToTopButton,
+  ] = useState(false);
+  const showScrollToTopButton = useCallback(() => {
+    if (modalRef.current === null) return;
+    const isScrollable =
+      modalRef.current.scrollHeight > modalRef.current.offsetHeight;
+    setShouldShowScrollToTopButton(isScrollable);
+    setIsFixed(false);
+  }, [modalRef, setShouldShowScrollToTopButton]);
```
abykovssc, line 19: you can omit `setShouldShowScrollToTopButton` dependency, according to official docs: https://reactjs.org/docs/hooks-reference.html#usestate

abykovssc, line 19: Turns out depending on refs isn't working as well, since ref is a mutable object, which exists for whole lifecycle of the component

ajkl2533: any suggestion?

src/components/FullscreenModal/Footer/useStickyFooter.ts
```diff
@@ -0,0 +1,75 @@
...
+  }, [isFixed, setIsFixed, modalRef, modalFooterRef]);
+  const [observer] = useState(new MutationObserver(() => isInView()));
+
+  useEffect(() => {
+    if (
+      modalRef.current !== null &&
+      modalFooterRef.current !== null &&
+      shouldShowScrollToTopButton
+    ) {
+      observer.observe(modalRef.current, {
```
abykovssc, line 50: I think `observer.disconnect` should be always called before calling `observer.observe` here. The thing is, body of this `useEffect` can be called multiple due to changes to dependency, therefore will will call observe multiple times, but will call `observe.disconnect` only once (when component is unmounted). Same applies to the listeners below, old ones has to be removed before new ones are added

src/components/FullscreenModal/Footer/useStickyFooter.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    if (isFixed && scrollOffset >= contentHeight) {
+      setIsFixed(false);
+    } else if (
+      !isFixed &&
+      scrollOffset < contentHeight - contentHeight * 0.001
+    ) {
+      setIsFixed(true);
+    }
+  }, [isFixed, setIsFixed, modalRef, modalFooterRef]);
```
abykovssc, line 41: Should depend only on `isFixed`, other dependencies aren't changing during lifecycle of the component

src/components/FullscreenModal/Footer/useStickyFooter.ts
```diff
@@ -0,0 +1,75 @@
...
+  const [observer] = useState(new MutationObserver(() => isInView()));
```
abykovssc, line 42: I kinda doubt MutationObserver should be put into state. I think proper approach would be creating a ref and storing observer in there, e.g. :
`const observerRef = useRef(null)`

```useEffect(()=>{
     if (observerRef.current !== null) {
         observerRef.current.disconnect();
     }
     observerRef.current = new MutationObserver(() => isInView());
     ...
})```

The difference is that MutationObserver is always using correct instance of `isInView()`

File: src/components/FullscreenModal/Footer/useStickyFooter.ts (hunk 3 above)
Comment (abykovssc) [reply src/components/FullscreenModal/Footer/useStickyFooter.ts:19]: don't depend on it, just remove it from the list of dependencies. If you need to perform update whenever dom element changes, there is something called callbacks refs: https://reactjs.org/docs/refs-and-the-dom.html#callback-refs

Comment (abykovssc) [reply src/components/FullscreenModal/Footer/useStickyFooter.ts:19]: But actually you can simply use empty array for dependencies, or don't use `useCallback` here at all. Everything that `showScrollToTopButton` operates on is guaranteed to be the same between renders 🤷

File: src/components/FullscreenModal/Footer/useStickyFooter.ts [hunk 7]
```diff
@@ -0,0 +1,75 @@
...
+    const isScrollable =
+      modalRef.current.scrollHeight > modalRef.current.offsetHeight;
+    setShouldShowScrollToTopButton(isScrollable);
+    setIsFixed(false);
+  }, [modalRef, setShouldShowScrollToTopButton]);
+
+  useEffect(() => {
+    showScrollToTopButton();
+    window.addEventListener('resize', debounce(showScrollToTopButton, 200));
+  }, [modalRef, showScrollToTopButton]);
```
abykovssc, line 24: I think both dependencies can be dropped (they both don't change between the renders)

src/components/FullscreenModal/Footer/useStickyFooter.ts
```diff
@@ -0,0 +1,75 @@
...
+    }
+    return () => {
+      observer.disconnect();
+    };
+  }, [
+    observer,
+    modalRef,
+    modalFooterRef,
+    shouldShowScrollToTopButton,
+    isInView,
```
abykovssc, line 68: Only `shouldShowScrollToTopButton` and `isInView` should be and dependencies, the rest doesn't really change between the renders

src/components/FullscreenModal/Header/useStickyHeader.ts
```diff
@@ -0,0 +1,43 @@
...
+    if (modalRef.current === null || modalHeaderRef.current === null) return;
+    const scrollOffset = modalRef.current.scrollTop;
+    const headerHeight = modalHeaderRef.current.scrollHeight;
+
+    if (isFixed && scrollOffset <= headerHeight) {
+      setIsFixed(false);
+    } else if (!isFixed && scrollOffset > headerHeight) {
+      setIsFixed(true);
+    }
+  }, [isFixed, setIsFixed, modalRef, modalHeaderRef]);
```
abykovssc, line 20: `isFixed` can be the only dependency

src/components/FullscreenModal/Header/useStickyHeader.ts
```diff
@@ -0,0 +1,43 @@
...
+  const [observer] = useState(new MutationObserver(() => isInView()));
```
abykovssc, line 21: For this block the same thing as `useStickyHeader` works, e.g. use `useRef` to keep mutation observer instance, disconnect old mutation observers whenever the useEffect body is executed, and clean old listener on unmount and useEffect executions

src/components/FullscreenModal/Header/useStickyHeader.ts
```diff
@@ -0,0 +1,43 @@
...
+      });
+
+      isInView();
+      modalRef.current.addEventListener('scroll', debounce(isInView, 100));
+      window.addEventListener('resize', debounce(isInView, 200));
+    }
+    return () => {
+      observer.disconnect();
+    };
+  }, [observer, modalRef, modalHeaderRef, isInView]);
```
abykovssc, line 38: `isInView` is the only changing dependency, the rest doesn't seem to change between renders

src/components/FullscreenModal/FullscreenModal.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { FlexContainer } from '../FlexContainer';
+import ModalContext from './Context/Context';
+import ModalHeader from './Header/Header';
+import ModalContent from './Content/Content';
+import ModalSidebar from './Sidebar/Sidebar';
+import ModalFooter from './Footer/Footer';
+import { FullscreenModalSizes } from './FullscreenModal.enums';
+import { FullscreenModalProps } from './FullscreenModal.types';
+
+export interface FullscreenModalSubcomponents {
```
abykovssc, line 15: shouldn't this interface be in `FullscreenModal.types.ts` file?

ajkl2533: It would create dependency cycle when subcomponents are imported in `.types.ts`

src/components/FullscreenModal/hooks/useStickyFooter.ts
```diff
@@ -16,12 +18,21 @@ export const useStickyFooter = (
...
 
   useEffect(() => {
     showScrollToTopButton();
     window.addEventListener('resize', debounce(showScrollToTopButton, 200));
-  }, [modalRef, showScrollToTopButton]);
+
+    return () => {
+      window.removeEventListener(
+        'resize',
+        debounce(showScrollToTopButton, 200),
```
abykovssc, line 31: i think `removeEventListener` removes by instance,  and `debounce` returns new function each time 🤔 
E.g. should be something like this:
```
const listener = debounce(showScrollToTopButton, 200);
window.addEventListener('resize', listener);
 return () => {
      window.removeEventListener(
        'resize',
        listener)
```

src/components/FullscreenModal/hooks/useStickyObserver.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export const useStickyObserver = (
+  modalRef: React.MutableRefObject<HTMLElement>,
+  elementRef: React.MutableRefObject<HTMLElement>,
+  isInView: () => void,
+): void => {
+  const observerRef = useRef(null);
+  const initialRender = useRef(true);
+  const scrollListener = debounce(isInView, 100);
+  const resizeListener = debounce(isInView, 200);
```
abykovssc, line 12: This approach for listeners is valid in useStickyFooter, but not going work here, since useEffect below will be run multiple times, so second time it will launched, the references to the listeners will be lost. So I'd use ref here instead and create listener in `useEffect`. Example:
```
const listeners = useRef(null);
...
useEffect(()=>{
    if(listeners.current != null) {
        currentModalRef.removeEventListener('scroll', listeners.current.scrollListener);
        window.removeEventListener('resize', listeners.current.resizeListener);
    }
    listeners.current = {resizeListener: debounce(isInView, 200), scrollListener:debounce(isInView, 100)};
    currentModalRef.addEventListener('scroll', listeners.current.scrollListener);
    window.addEventListener('resize',  listeners.current.resizeListener);    
});
```

abykovssc (approved):
LGTM, but one last thing to consider

src/components/FullscreenModal/FullscreenModal.tsx
```diff
@@ -0,0 +1,121 @@
...
+  const hasLayoutSidebar = sidebarConfig[0] > 0;
+  const totalContentWidth = contentConfig[0] + sidebarConfig[0];
+
+  if (hasLayoutSidebar && isUndefined(sidebar)) {
+    // eslint-disable-next-line no-console
+    console.warn(
+      `You chose to use modal layout with sidebar (current: ${layout}) but you didn't provide sidebar content.
+You should either provide content in "sidebar" property or switch layout to "${FullscreenModalLayouts.single6}" or "${FullscreenModalLayouts.single8}"
+`,
+    );
```
benjamine, line 72: could we throw an Error here? (if this isn't expected usage, I would rather not allow it)

src/components/FullscreenModal/FullscreenModal.tsx
```diff
@@ -0,0 +1,121 @@
+import React, { useRef } from 'react';
```
benjamine, line 1: is there for example a README.md next to these, to explain what is a full-screen modal and why/when should I use it? (part of storybook?)

ajkl2533: > is there for example a README.md next to these
 There is the Storybook file with all supported cases and you can see source code directly in Storybook Docs
 
 > why/when should I use it?
 That is more for UX team, they are aware when to use this pattern
 
But yes we should improve documentation of the components, but as that is work for weeks and it should be probably done in cooperation with UX, if this doesn't get some priority (and TBH it probably will not get any) I'd rather stay with the documentation provided automatically by SB.
I tried to do my best with documenting props and possible values in SB as well as adding "playground" stories (you can try combination of props directly in SB) to components where it makes sense.

src/components/FullscreenModal/FullscreenModal.tsx
ajkl2533: So I was thinking about this and I think that throwing an error is too strong since this isn't a violation which could break the application, I personally would leave throwing for things which could cause actual errors in the application.
However, I did change `console.warn` to `console.error`.

src/components/FullscreenModal/FullscreenModal.tsx
```diff
@@ -0,0 +1,121 @@
...
+import { getColor, getDepth } from '../../utils/helpers';
+import ModalHeader from './Header/Header';
+import ModalFooter from './Footer/Footer';
+import { FullscreenModalLayouts } from './FullscreenModal.enums';
+import { ColumnConfigMap, FullscreenModalProps } from './FullscreenModal.types';
+import { Col, Container, Row } from '../layout';
+
+const BaseModal = styled.div`
+  width: 100vw;
+  height: 100vh;
```
vaclav-zeman, line 15: Nitpick but since it's fixed, plain old % would be good enough

src/components/FullscreenModal/FullscreenModal.tsx
vaclav-zeman: If the modal is passed `layout={.."sidebar"}` but the comp is not supplied in the `sidebar` prop -> that should throw Error in my opinion as well as it sounds like a user mistake. Console logs/errors get lost easily

src/components/FullscreenModal/FullscreenModal.tsx
```diff
@@ -0,0 +1,121 @@
...
+  header,
+  content,
+  sidebar,
+  footer,
+  onClose = noop,
+}) => {
+  const modalRef = useRef<HTMLDivElement>(null);
+
+  const {
+    header: headerConfig,
```
vaclav-zeman, line 58: readability suggestion - you could use `header: [headerCols, headerOffset]` etc instead of indexes

src/components/FullscreenModal/Header/Header.tsx
```diff
@@ -0,0 +1,96 @@
...
+  offset,
+  modalRef,
+  handleClose,
+}) => {
+  const modalHeaderRef = useRef(null);
+  const { isFixed } = useStickyHeader(modalRef, modalHeaderRef);
+
+  return (
+    <>
+      {isFixed && (
```
vaclav-zeman, line 45: When it's fixed, why are both headers displayed?

src/components/FullscreenModal/ScrollToTop/ScrollToTop.tsx
```diff
@@ -0,0 +1,51 @@
...
+  padding: 0;
+  text-align: center;
+  line-height: ${pxToRem(13)};
+`;
+
+const ButtonText = styled.span`
+  margin-top: ${pxToRem(4)};
+`;
+
+const ScrollToTop: React.FC<ScrollToTopProps> = ({ onClick = undefined }) => {
```
vaclav-zeman, line 26: - nitpick but this should be named `ScrollToTopButton`
- unnecessary default value?

src/components/FullscreenModal/ScrollToTop/ScrollToTop.tsx
```diff
@@ -0,0 +1,51 @@
...
+  };
+
+  return (
+    <ScrollToTopButton
+      color={ButtonColors.secondary}
+      variant={ButtonVariants.text}
+      onClick={handleClick}
+    >
+      <Icon name={SSCIconNames.arrowUp} />
+      <ButtonText>Scroll to top</ButtonText>
```
vaclav-zeman, line 42: I wonder how to deal with hardcoded texts here - since it's a design system, this should allow custom children that can be translatable 🤔

src/components/FullscreenModal/hooks/useStickyFooter.ts
```diff
@@ -0,0 +1,61 @@
...
+    const scrollOffset =
+      modalRef.current.scrollTop + modalRef.current.offsetHeight;
+    const contentHeight =
+      modalRef.current.scrollHeight - modalFooterRef.current.scrollHeight;
+
+    if (isFixed && scrollOffset >= contentHeight) {
+      setIsFixed(false);
+    } else if (
+      !isFixed &&
+      scrollOffset < contentHeight - contentHeight * 0.001
```
vaclav-zeman, line 48: magic const? 😄

src/components/FullscreenModal/hooks/useStickyObserver.ts
```diff
@@ -0,0 +1,59 @@
...
+    return () => {
+      observerRef.current.disconnect();
+      currentModalRef.removeEventListener(
+        'scroll',
+        listeners.current.scrollListener,
+      );
+      window.removeEventListener('resize', listeners.current.resizeListener);
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [isInView]);
```
vaclav-zeman, line 58: Does it work properly when `modalRef.current` is empty on first render? Seems like it wouldn't re-run the effect but it probably does thanks to some other cond?

vaclav-zeman (commented):
Looks good, wish it could go without all the ref passing and complex observers but I don't have any better solution to offer

src/components/FullscreenModal/Header/Header.tsx
ajkl2533: I tried to do this with just one header displayed at a time, but it's making a mess when the one should be displayed and the second one hidden. Because it will change the`modalRef. scrollHeight` and it usually caused an infinite loop when the second footer gets displayed and the condition in hook get change which causes the second footer gets hidden and again and again :)

src/components/FullscreenModal/ScrollToTop/ScrollToTop.tsx
ajkl2533: we can do a prop for modal with a default value, good catch

src/components/FullscreenModal/hooks/useStickyFooter.ts
ajkl2533: `contentHeight * 0.001` is the threshold which tells when the footer gets sticky, I can extract that into const to be more explicit

src/components/FullscreenModal/hooks/useStickyObserver.ts
ajkl2533: Yep, it should be working, I advise going through previous @abykovssc comments. As I understand a ref is a mutable object which exists through the whole component lifecycle and thus it doesn't need to be specified as a dependency here.

src/components/FullscreenModal/hooks/useStickyFooter.ts
vaclav-zeman: Yep, that would be helpful

src/components/FullscreenModal/hooks/useStickyObserver.ts
vaclav-zeman: Yeah, that's true but my point is that ref.current is empty on first render (assuming it's ref attached to a JSX comp, not just value). But probably isInView is triggered after the first render so it works 🤔

vaclav-zeman: Ok, I see https://github.com/facebook/react/issues/14387#issuecomment-503616820 . In the past I think refs attached to JSX elements were sometimes null for the first render but maybe that's not an issue anymore

src/components/FullscreenModal/FullscreenModal.enums.ts
```diff
@@ -0,0 +1,12 @@
+export const FullscreenModalSizes = {
+  lg: 'lg',
+  md: 'md',
+  sm: 'sm',
+} as const;
+
+export const FullscreenModalLayouts = {
+  single6: 'single-6',
+  single8: 'single-8',
+  sidebar46: 'sidebar-4-6',
```
youssefssc, line 10: I see only one layout in the design specs with sidebar. Are the designs not complete and this is a direct request from @MateCziner ? Also the 10 columns layout is not in sync with the 12 columns global Grid. Can you please tell more about this decision?

ajkl2533: @youssefssc these layout was requested by @MateCziner and are not implemented in Figma. Not sure what you mean by: `Also the 10 columns layout is not in sync with the 12 columns global Grid`

youssefssc: Nevermind the second comment. I thought the 4-6 is a new 10 columns layout but found it's actually 12 cols and the sidebar is just offseted by 1 col

youssefssc (approved):
nice work LGTM

src/components/FullscreenModal/FullscreenModal.tsx
ajkl2533: Ok guys change to throw error :)

vaclav-zeman (approved):
LGTM

PR #23, opened 2020-12-16
feat(grid): add grid components

ajkl2533:


src/components/layout/Col/Col.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { ColProps, Cols } from './Col.types';
+
+const getColWidth = (cols: Cols): { flex: string } | { width: number } => {
+  if (cols === 'auto') return { flex: '1 1 auto' };
+
+  return { width: cols / 12 };
+};
+
+const Col: React.FC<ColProps> = ({ children, cols, offset }) => (
+  <Box ml={`${(100 / 12) * offset}%`} px={pxToRem(15)} {...getColWidth(cols)}>
```
vaclav-zeman, line 15: Imo it'd make sense for gutter (15) to be set in theme as well. Any app could then override it by their own theme

src/theme/widths.ts
```diff
@@ -0,0 +1,7 @@
+import { Widths } from './widths.types';
+
+const widths: Widths = {
```
vaclav-zeman, line 3: I'd rather rename this to `grid/layout.ts` as it'd be more generic - usable for `reportWidth, layoutWidth, gutter` etc

vaclav-zeman (approved):
LGTM, few suggestions

PR #28, opened 2021-01-08
feat(datatable): add ElementCounter component

ajkl2533:
Closes FEP-1527


src/components/Datatable/ElementCounter/ElementCounter.types.ts
```diff
@@ -0,0 +1,5 @@
+export interface ElementCounterProps {
+  totalLength: number;
```
youssefssc, line 2: what about the case when the numbers are large? in the design for example the total is rendered as "1.05M" .

youssefssc: ok @ajkl2533 I see you are using `formatNumber` and it works but renders 1.05m with small "m"

PrincAm: maybe adding of this use case to storybook would be helpful

PrincAm (approved):
LGTM

src/components/Datatable/ElementCounter/ElementCounter.types.ts
ajkl2533: sure guys :) fixed the uppercase and added abbreviated counts into SB

PR #30, opened 2021-01-12
Tabs update

vaclav-zeman:
Tabs refactored to allow more flexible API - heavily inspired by https://material-ui.com/components/tabs/


src/components/Tabs/Tab.tsx
```diff
@@ -0,0 +1,58 @@
...
+  font-size: ${pxToRem(18)};
+  line-height: ${pxToRem(20)};
+  padding-bottom: ${pxToRem(3)};
+  cursor: pointer;
+  border-bottom: 2px solid
+    ${({ $isSelected, color }) =>
+      $isSelected ? getColor(color) : getColor('graphiteHB')};
+
+  &:hover {
+    border-bottom: 2px solid ${({ color }) => color};
```
ajkl2533, line 21: shouldn't it be `getColor(color)`?

vaclav-zeman: yep, missed that

src/components/Tabs/Tab.tsx
```diff
@@ -0,0 +1,58 @@
...
+  onClick,
+  color = ColorTypes.blueberryClassic,
+  value,
+}) => {
+  const isLink = value?.toString()?.startsWith('/');
+  const handler = isLink ? { to: value } : { onClick: () => onClick(value) };
+
+  return (
+    <Label
+      $isSelected={isSelected}
```
ajkl2533, line 40: why prefix with `$` sign?

vaclav-zeman: transient props, they will not be passed to rendered DOM so no more errors 🎉

vaclav-zeman: https://styled-components.com/docs/api#transient-props

src/components/Tabs/Tab.tsx
```diff
@@ -0,0 +1,58 @@
...
+      color={color}
+      {...handler}
+    >
+      {children}
+    </Label>
+  );
+};
+
+Tab.propTypes = {
+  children: PropTypes.oneOfType([PropTypes.string, PropTypes.node]).isRequired,
```
ajkl2533, line 51:
```suggestion
  children: PropTypes.node.isRequired,
```
`PropTypes.string` is already included in `PropTypes.node`

src/components/Tabs/Tab.tsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
```
ajkl2533, line 1: This component is not exported from Design system library. You should add export to https://github.com/securityscorecard/design-system/blob/tabs-refactor/src/components/Tabs/index.ts

src/components/Tabs/Tab.tsx
ajkl2533: cool, didn't know that :)

src/components/Tabs/Tabs.types.ts
```diff
@@ -1,19 +1,19 @@
...
+  $isSelected: boolean;
   color: string;
 }
-
-export interface Tab {
-  title: string;
-  color: string;
-  Component: React.ReactNode;
+export interface TabProps {
+  children: string | React.ReactNode;
```
ajkl2533, line 8:
```suggestion
  children: React.ReactNode;
```

ajkl2533 (approved):
LGTM, once suggestions are fixed

PR #35, opened 2021-01-13
feat(Datatable): add BatchActions component

ajkl2533:
Closes FEP-1528


src/components/Datatable/BatchActions/BatchActions.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { Dropdown } from '../Dropdown';
+import { Icon } from '../../Icon';
+import { SSCIconNames } from '../../Icon/Icon.enums';
+import { ColorTypes } from '../../../theme/colors.enums';
+
+const BatchActionButton = styled(Button).attrs((props) => ({
+  variant: ButtonVariants.text,
+  ...props,
+}))`
+  ${createPaddingSpacing({ horizontal: 0.8 })}
```
PrincAm, line 26: for my better understanding, when should I use `createPaddingSpacing` instead of defining padding directly?

src/components/Datatable/Dropdown/DropdownPane.tsx
```diff
@@ -0,0 +1,103 @@
...
+            RouterLink = requireRouterLink();
+          }
+
+          const domTag = isNotUndefined((action as AbsoluteLinkActionKind).href)
+            ? 'a' // render 'a' tag if 'href' is present
+            : isNotUndefined((action as RelativeLinkActionKind).to)
+            ? RouterLink // render 'Link' if 'to' is present
+            : undefined; // use default
+
+          if (
```
PrincAm, line 72: pls, what is this use case? `to` is defined but you don't have `react-router-dom` dependency?

ajkl2533: since `RouterLink` is imported dynamically only when needed, there can be some error and it's better to not render anything than crash app

PrincAm (approved):
LGTM

src/components/Datatable/BatchActions/BatchActions.tsx
ajkl2533: Great question but TBH I have no clear answer for that. I hope we can deal with this dilemma once we introduce spacing scales. For now I actually would go with using `create...Spacing` functions only when you want to pass `padding` and/or `margin` prop in the component. I'll remove it from here.

PR #36, opened 2021-01-14
feat(Filters): add FilterRow component

PrincAm:
Implementation of filters row. Is highly probable that API of component(s) will change when adding/removing filters logic will be implemented. For now, it is dumb.


src/components/Filters/Filters.tsx
```diff
@@ -0,0 +1,29 @@
...
+  <FlexContainer flexDirection="column">
+    {options.map(({ uuid, ...rest }, index) => (
+      <FilterRow key={uuid} isFirstRow={index === 0} uuid={uuid} {...rest} />
+    ))}
+  </FlexContainer>
+);
+
+export default Filters;
+
+const RowPropType = PropTypes.exact({
```
ajkl2533, line 19: Will this be reused somewhere? If yes please move it to `Filters.types.ts` if not put it directly in `Filters.propTypes`

src/components/Filters/Filters.tsx
```diff
@@ -0,0 +1,29 @@
...
+  uuid: PropTypes.string.isRequired,
```
ajkl2533, line 20: I wouldn't call it `uuid` since `uuid` have strict format. I would call it `name` or just `id`.

src/components/Filters/Filters.types.ts
```diff
@@ -0,0 +1,13 @@
+import { Option } from './Select/Select.types';
+
+export interface FilterRow {
+  uuid: string;
+  conditionOptions: Option[];
+  dataOptions: Option[];
```
ajkl2533, line 6: Hmm I think that `dataOption` should be shared across all filter rows. As it should represent columns in table. Not sure if filters are used without the datatable does make sense to have different data points for different rows.

src/components/Filters/Filters.types.ts
```diff
@@ -0,0 +1,13 @@
+import { Option } from './Select/Select.types';
+
+export interface FilterRow {
+  uuid: string;
+  conditionOptions: Option[];
```
ajkl2533, line 5: I'm checking with @MateCziner if this has to be this flexible if we can just use some list with posible input value types to determine what conditions should be available.

My original idea was to provide someting like:
```ts
{
  name: string;
  dataPoint: string; // <- table column identifier
  valueType: string | number | datetime | ...; // what can be used as value type
  onRemove: () => void;
  onAdd?: () => void; // we can use it for event tracking
  isFilterApplied?: boolean;
}

File: src/components/Filters/StateButton/StateButton.types.ts [hunk 5]
```diff
@@ -0,0 +1,5 @@
+export interface StateButtonProps {
+  uuid: string;
+  onClick: (uuid: string) => void;
```
ajkl2533, line 3: Again `uuid`

src/components/Filters/StateButton/StateButton.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { StateButtonProps } from './StateButton.types';
+import { useStateButtonIcon } from '../hooks/useStateButton';
+
+const Popup = styled(FlexContainer)`
+  position: absolute;
+  visibility: hidden;
+  z-index: ${getDepth('tooltip')};
+  padding: ${pxToRem(8.5, 0, 8.5, 10)};
+  background: ${getColor('strawberry')};
+  border-radius: ${getBorderRadius} 0 0 ${getBorderRadius};
```
ajkl2533, line 24: I'm not sure if this wouldn't be clashing with the namespacing babel plugin in platform. I would rather split this to `border-top-left-radius` and `border-bottom-left-radius` just to be sure.

src/components/Filters/StateButton/StateButton.tsx
```diff
@@ -0,0 +1,90 @@
...
+  transform: translateY(-50%);
+`;
+
+const RemoveButton = styled(FlexContainer)`
+  position: relative;
+  padding: ${pxToRem(8, 5.5)};
+  cursor: pointer;
+  &:hover {
+    background-color: ${getColor('strawberry')};
+    border-radius: 0 ${getBorderRadius} ${getBorderRadius} 0;
```
ajkl2533, line 37: Same as above

src/components/Filters/StateButton/StateButton.tsx
```diff
@@ -0,0 +1,90 @@
...
+  }
+  &:hover ${Popup} {
+    visibility: visible;
+    opacity: 1;
+  }
+`;
+
+const LightText = styled(Text)`
+  color: ${getColor('graphite5H')};
+  line-height: ${pxToRem(15)};
```
ajkl2533, line 47: `${getLineHeight('md')}

src/components/Filters/StateButton/StateButton.tsx
```diff
@@ -0,0 +1,90 @@
...
+  font-weight: 500;
```
ajkl2533, line 48: `${getFontWeight('medium')}`

src/components/Filters/hooks/useStateButton.ts
```diff
@@ -0,0 +1,40 @@
...
+  const [iconColor, setIconColor] = useState<Color>(timesIconColor);
+
+  useEffect(() => {
+    if (isFilterApplied) {
+      setIconName(SSCIconNames.check);
+      setIconColor(checkIconColor);
+    }
+  }, [isFilterApplied]);
+
+  const handleMouseOut = () => {
```
ajkl2533, line 25: These `handleMouseOut` and `handleMouseOver` methods probably should be wrapped in `useCallback` since you are passing them into `RemoveButton` component. @abykovssc could you please confirm I'm still not 100% sure how this should be used :D

src/components/Filters/StateButton/StateButton.tsx
```diff
@@ -0,0 +1,90 @@
...
+  padding: ${pxToRem(8.5, 0, 8.5, 10)};
+  background: ${getColor('strawberry')};
+  border-radius: ${getBorderRadius} 0 0 ${getBorderRadius};
+  opacity: 0;
+  right: calc(100%);
+  top: 50%;
+  transform: translateY(-50%);
+`;
+
+const RemoveButton = styled(FlexContainer)`
```
ajkl2533, line 31: I would probably use `styled.button` directly as set flexbox properties. It looks odd when you use FlexContainer as button

src/components/Filters/FilterRow/FilterRow.tsx
```diff
@@ -0,0 +1,75 @@
...
+import operatorOptions from '../data/operator-options.json';
+
+const Container = styled(FlexContainer)`
+  margin-bottom: ${pxToRem(8)};
+  &:last-of-type {
+    margin-bottom: ${pxToRem(0)};
+  }
+`;
+
+const SplitField = styled.div`
```
ajkl2533, line 22: if you do 
```js
const SplitField = styled.div<SplitFieldProps>
```
you don't have to pass type annotation in the `width` property

src/components/Filters/hooks/useStateButton.ts
abykovssc: No need to wrap it imo, since this hook is used in one place only (as i understand it isn't some kind of public api), and RemoveButton isn't memoized, so wrapping this handlers into `useCallback` isn't going to achieve anything

src/components/Filters/Filters.tsx
PrincAm: I pick `id` to have it clear it should be unique for further remove handling

src/components/Filters/StateButton/StateButton.tsx
PrincAm: Yep, it is weird. It was my init version then I changed to FlexContainer. Will use `button`

src/components/Filters/FilterRow/FilterRow.tsx
PrincAm: 👍 Good to keep types syntax consistent

src/components/Filters/Filters.types.ts
PrincAm: It makes sense to move it one level up

src/components/Filters/Filters.enums.ts
```diff
@@ -0,0 +1,10 @@
+export const InputTypes = {
+  string: 'string',
+  multipleString: 'multipleString',
+  uint: 'uint',
```
ajkl2533, line 4: I would rename it to `count`, no need to bring here terms like `unsigned integer` :D

src/components/Filters/Filters.enums.ts
```diff
@@ -0,0 +1,10 @@
...
+  int: 'int',
```
ajkl2533, line 5: `number` instead of `int`

src/components/Filters/Filters.enums.ts
```diff
@@ -0,0 +1,10 @@
...
+  enum: 'enum',
+  multipleEnum: 'multipleEnum',
+  date: 'date',
+  timeInterval: 'timeInterval',
```
ajkl2533, line 9: `dateInterval` would be probably better since we probably want to select days and not actual timeframe

src/components/Filters/Filters.tsx
```diff
@@ -5,25 +5,36 @@ import { FlexContainer } from '../FlexContainer';
...
     ))}
   </FlexContainer>
 );
 
 export default Filters;
 
 Filters.propTypes = {
-  options: PropTypes.arrayOf(
+  dataOptions: PropTypes.arrayOf(OptionPropType).isRequired,
+  rows: PropTypes.arrayOf(
```
ajkl2533, line 28: I'm missing the connection between table columns and conditions + inputTypes. 
e.g. I will select `col x` and how do the filters know that it should show me `string` input and not `enum`?

PrincAm: Hm, good catch, it is not there. I will probably move input type to data options to get this mapping

PrincAm: There has to dependency Data point -> Condition -> Input type

ajkl2533: @PrincAm I know I'm iterating this over and over but what about:
```js
interface FiltersProps {
  rows: {
    [key: dataPoint]: { // <-- column identifier
      conditions: [], // <-- list of conditions
      defaultCondition: '',
      inputType: '' // <-- this should probably accept React components also
    }
  }
}
```

ajkl2533:
```const options = {
  dataPoint: {
    // issues count
    conditions: [
      {
        id: string,
        input:
          InputValueTypes |
          React.Element |
          {
            type: InputValueTypes | React.Element,
            props: { options },
          },
        isDefault: bool, // true
        label: string, // is at least
        value: string, // value
        operation: string,
      },
    ],
  },
};

const defaultRows = {
  operator: 'and', // 'and' | 'or'
  filters: {
    value: '',
    dataPoint: '',
    condition: '',
  },
};
// is, is not, contains, does not contain
const issuesCount = {
  conditions: [
    {
      label: 'greater than',
      value: 'gt',
      operation: 'gt'
      input: {
        type: CountFilter,
        props: {
          min: 10,
          max: 100,
        }
      },
    },
  ],
};

Comment (PrincAm) [reply src/components/Filters/Filters.tsx]: @ajkl2533 One more round 🙂 API was updated as we agreed. Logic of handling between selectors and input types within `FiltersRow` will be in next PR.

File: src/components/Filters/Filters.tsx [hunk 17]
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { FlexContainer } from '../FlexContainer';
+import { FilterRow } from './FilterRow';
+import { DataPointPropTypes, FiltersProps } from './Filters.types';
+import { Operators } from './Filters.enums';
+
+const generateId = ({ operator, dataPoint, condition, input }, index) =>
+  `${operator}-${dataPoint}-${condition}-${input}-${index}`;
```
PrincAm, line 10: Not perfect but it seems to me be better than let user define id and add another property. Props values should be `string` or `undefined`.

src/components/Filters/Filters.types.ts
```diff
@@ -0,0 +1,71 @@
...
+  dataPoints: DataPoint[];
+  value: Filter[];
+  onApply: (filters: Filter[]) => void;
+  onChange?: (filters: Filter[]) => void;
+}
+
+export const DataPointPropTypes = PropTypes.exact({
+  conditions: PropTypes.arrayOf(
+    PropTypes.exact({
+      input: PropTypes.oneOf([
```
ajkl2533, line 51: This should be `oneOfType`. `oneOf` is for exact values

src/components/Filters/Filters.types.ts
```diff
@@ -0,0 +1,71 @@
...
+export interface Filter {
+  operator: OperatorTypes;
+  dataPoint: string;
+  condition: string;
+  input: string;
+}
+
+export interface FiltersProps {
+  dataPoints: DataPoint[];
+  value: Filter[];
```
ajkl2533, line 43: I'm not sure about the propName `value`. One thing is that this is an array, so, the plural is probably more appropriate, like `values`. The other thing is the word itself, maybe `filtersState` or simple `data` would be better

src/components/Filters/FilterRow/FilterRow.types.ts
```diff
@@ -0,0 +1,11 @@
+import { DataPoint, Filter } from '../Filters.types';
+
+export interface SplitFieldProps {
+  width?: number;
+}
+
+export interface FilterRowProps extends Filter {
+  id: string;
+  dataPoints: DataPoint[];
+  isFirstRow?: boolean;
```
ajkl2533, line 10: this prop isn't defined in propTypes and is not even used in the component

PrincAm: good catch! i missed that, will remove it

src/components/Filters/Filters.types.ts
PrincAm: I took the name from KendoReact Filter but understand to your concern. I will use `data` as more generic and shorter name

src/components/Filters/WhereOption/WhereOption.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+import { Text } from '../../typography';
+import { TextSizes } from '../../typography/Text/Text.enums';
+import { FlexContainer } from '../../FlexContainer';
+import { getBorderRadius, getColor, pxToRem } from '../../../utils/helpers';
+
+const Container = styled(FlexContainer)`
+  background: ${getColor('graphite3H')};
+  border-radius: ${getBorderRadius};
+  width: '100%';
```
ajkl2533, line 12: hm those quotes shouldn't be necessary

PrincAm: oh, good catch! it turned out `width` doesn't have to be declared at all. removed

ajkl2533 (approved):

PR #38, opened 2021-01-18
feat(Datatable): add ToolsTabItems

youssefssc:
closes FEP-1524


src/components/Datatable/ToolsTabItem/ToolsTabItem.tsx
```diff
@@ -0,0 +1,160 @@
...
+import { IconTypes, SSCIconNames } from '../../Icon/Icon.enums';
+import { SSCIcons, Types } from '../../Icon/Icon.types';
+import Button from '../../Button/Button';
+import StyledIcon from '../../Button/StyledIcon';
+import { ButtonSizes } from '../../Button/Button.enums';
+import { ToolsTabItemProps } from './ToolsTabItem.types';
+import { ToolsTabItemStates } from './ToolsTabItem.enum';
+
+const DefaultToolsTabItem = styled(Button)`
+  &&& {
```
ajkl2533, line 17: This shouldn't be needed

src/components/Datatable/ToolsTabItem/ToolsTabItem.types.ts
```diff
@@ -0,0 +1,12 @@
+import { ButtonProps } from '../../Button/Button.types';
+import { ToolsTabItemStates } from './ToolsTabItem.enum';
+
+export type ItemStates = typeof ToolsTabItemStates[keyof typeof ToolsTabItemStates];
+
+export interface ToolsTabItemProps extends ButtonProps {
+  itemState?: ItemStates;
+  activateTool: React.MouseEventHandler;
```
ajkl2533, line 8: It's better to start handlers with `on` prefix to distinguish it with data props:
- `onToolActivate`
- `onToolDeactivate`

src/components/Datatable/ToolsTabItem/ToolsTabItem.tsx
```diff
@@ -0,0 +1,160 @@
...
+  }
+`;
+
+const ActiveToolsTabItem = styled(Button)`
+  &&& {
+    padding: ${pxToRem(6, 12)};
+    background-color: ${getColor('graphite5H')};
+    color: ${getColor('blueberryClassic')};
+    border-radius: ${pxToRem(20)};
+    border-color: ${getColor('blueberryClassic')};
```
ajkl2533, line 39: `border: 2px solid ${getColor('blueberryClassic')}`?

src/components/Datatable/ToolsTabItem/ToolsTabItem.tsx
ajkl2533: I would actually recreate this from HTML `button` tag. You probably wont need all the functionality from `Button` component (like dynamic tag resolution). Since this will be used just for the tooltab where all tools will be defined by the `Datatable` component itself so it doesn't need to be that flexible. And if we decide later to add this variant in the `Button` component we would still need styled directly there.

src/components/Datatable/ToolsTabItem/ToolsTabItem.tsx
```diff
@@ -0,0 +1,160 @@
...
+    &.focus {
+      color: ${getColor('graphite4B')};
+      svg {
+        color: ${getColor('graphite2B')};
+      }
+    }
+  }
+`;
+
+const ActiveToolsTabItem = styled(Button)`
```
ajkl2533, line 33: Those styled should be probably inherited from the `DefaultToolsTabItem` not from the base `Button` component

src/components/Datatable/ToolsTabItem/ToolsTabItem.tsx
```diff
@@ -0,0 +1,160 @@
...
+  &&& {
+    padding: ${pxToRem(6, 12)};
```
ajkl2533, line 35: if you define the `border-width` in the default (base) component and set the `border-color` to transparent you won't need to do the magic with paddings

src/components/Datatable/ToolsTabItem/ToolsTabItem.tsx
```diff
@@ -0,0 +1,160 @@
...
+    activateTool();
+    setCurrentItemState(ToolsTabItemStates.active);
+  };
+
+  const onClickActive = () => {
+    deactivateTool();
+    setCurrentItemState(ToolsTabItemStates.applied);
+  };
+
+  const onClickApplied = () => {
```
ajkl2533, line 109: this is the same as `onClickDefault`

src/components/Datatable/ToolsTabItem/ToolsTabItem.tsx
youssefssc: This makes more sense, using html `button` instead now

src/components/Datatable/ToolsTabItem/ToolsTabItem.tsx
youssefssc: changed to `onClickActive` and `onClickNotActive`

src/components/Datatable/ToolsTabItem/ToolsTabItem.tsx
```diff
@@ -0,0 +1,157 @@
...
+  );
+
+  const commonProps = {
+    margin,
+    ...props,
+  };
+
+  return (
+    <>
+      {!isDisabled &&
```
ajkl2533, line 122: @youssefssc can we replace this with 
```js
if (isDisabled) {
  return null;
}

if (currentToolState === ToolsTabItemStates.applied) {
  return (...)
 }

if (currentToolState === ToolsTabItemStates.active) {
  return (...)
}

return (...)
```

it's more readable than nested ternaries

ajkl2533 (approved):
Please drop the revert commits from PR (not necessary) and one small comment

src/components/Datatable/ToolsTabItem/ToolsTabItem.tsx
youssefssc: it is nicer, done :)

PR #39, opened 2021-01-25
feat(FullscreenModal): add getModalRef and expose modalRef to users

ajkl2533:


PrincAm (approved):
Nice component. There is bunch of logic within custom hooks which I hope it works as expected. Code LGTM. GJ 👍

PR #40, opened 2021-01-26
feat(accordion): added accordion with state management

jurajhrib:


src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { AccordionProps } from './Accordion.types';
+import AccordionCollapsible from './AccordionCollapsible';
+
+const Accordion: React.FC<AccordionProps> = ({
+  isCollapsedOnOpen = true,
+  items,
+  ...props
+}) => {
+  const [openIndexes, setOpenIndexes] = useState(
+    items.filter((item) => item.isOpen).map((item, index) => index),
```
vaclav-zeman, line 14: I'm a bit wary of using indices as keys (since data can be sorted in the parent and issues happens) although there's assumption that Accordion will always get some constant data. 

I would still much rather add an `id` to the items array

src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,64 @@
...
+    >
+      {items.map((item, index) => (
+        <AccordionCollapsible
+          key={`accordion-wrapper-item-${item.title}`}
+          handleHeaderClick={handleClick}
+          index={index}
+          isOpen={openIndexes.includes(index)}
+          title={item.title}
+        >
+          {item.content} {isCollapsedOnOpen}
```
vaclav-zeman, line 46: Is this just rendering the boolean?

src/components/Accordion/AccordionCollapsible.tsx
```diff
@@ -0,0 +1,80 @@
...
+  className,
+  handleHeaderClick,
+  isOpen = false,
+  index,
+  title,
+}) => (
+  <Container className={className} isOpen={isOpen}>
+    <Header alignItems="center" onClick={() => handleHeaderClick(index)}>
+      <StyledIcon
+        isRotated={isOpen}
```
vaclav-zeman, line 58: Not sure what's @ajkl2533 take on that but I like the new transient props for this so you don't have to use shouldForwardProp - https://styled-components.com/docs/api#transient-props

src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,64 @@
...
+  );
+
+  const handleClick = useCallback(
+    (index: number) => {
+      setOpenIndexes(
+        openIndexes.includes(index)
+          ? openIndexes.filter((i) => i !== index)
+          : isCollapsedOnOpen
+          ? [index]
+          : [...openIndexes, index],
```
vaclav-zeman, line 24: Just my preference but nested ternaries are hell to read 😄

jurajhrib: Yeah, wondered if it's gonna catch someone's eyes. Can break it into sub call or something.

src/components/Accordion/Accordion.tsx
jurajhrib: @vaclav-zeman we can do it, but I was thinking that it kinda puts the burden to the component user and he will just snap numbers there all the same. But I get your point.

src/components/Accordion/Accordion.tsx
vaclav-zeman: I think it's used all over ui-core anyway so doesn't matter to me much

src/components/Accordion/Accordion.tsx
vaclav-zeman: Yep, trade off for sure. But it's fine if user adds numbers since they will stick to the item even after sorting/removing item

jurajhrib: @vaclav-zeman done 😌

src/components/Accordion/AccordionCollapsible.tsx
jurajhrib: @vaclav-zeman if our styled supports it and eslint won't scream at me for invalid property name, why not 😁

jurajhrib: @vaclav-zeman tried it, and it works fine, so depends on @ajkl2533 whether he likes it. It is a little bit odd to put special character into prop name. But it's internal stuff on the other hand, so it should not irritate anyone.

ajkl2533: I'm totally fine with using transient props for internal API. I would rather not to expose it to end-users tho.

src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,65 @@
...
+          title={item.title}
+        >
+          {item.content}
+        </AccordionCollapsible>
+      ))}
+    </FlexContainer>
+  );
+};
+
+const AccordionItemPropType = PropTypes.exact({
```
ajkl2533, line 53: There is not written convention to define specific PropType in the `*.types.ts` file and import it from there.

src/components/Accordion/Accordion.types.ts
```diff
@@ -0,0 +1,21 @@
+import { ReactNode } from 'react';
+
+export type AccordionItem = {
+  title: string;
+  content: ReactNode;
+  isOpen?: boolean;
+  id: string | number;
```
ajkl2533, line 7: I would probably extract id type like `export type AccordionItemId = string | number` and then reuse it since from the implementation I see that id in AccordionItem and AccordionCollapsiblePros are the same. You can also use it here https://github.com/securityscorecard/design-system/pull/40/files#diff-a31ce76dd91c59f1f5ac8d8adbe78ee1ca5b3e195f546b8419f6f80554f6d626R18.

src/components/Accordion/AccordionCollapsible.tsx
```diff
@@ -0,0 +1,80 @@
...
+const AccordionCollapsible: React.FC<AccordionCollapsibleProps> = ({
+  children,
+  className,
+  handleHeaderClick,
+  isOpen = false,
+  id,
+  title,
+}) => (
+  <Container className={className} isOpen={isOpen}>
+    <Header alignItems="center" onClick={() => handleHeaderClick(id)}>
```
ajkl2533, line 56: Since we should care little bit about accessibility you should follow:
- https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-static-element-interactions.md#case-this-element-acts-like-a-button-link-menuitem-etc
- https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/click-events-have-key-events.md

ajkl2533 (approved):
Couple of small issues, but looks good.

⚠️  Don't forget add export in https://github.com/securityscorecard/design-system/blob/main/src/components/index.ts

src/components/Accordion/Accordion.types.ts
```diff
@@ -0,0 +1,31 @@
...
+export interface AccordionProps {
+  isCollapsedOnOpen?: boolean;
+  items: AccordionItem[];
+}
+
+export const AccordionItemPropType = PropTypes.exact({
+  title: PropTypes.string.isRequired,
+  content: PropTypes.node.isRequired,
+  isOpen: PropTypes.bool,
+  id: PropTypes.oneOfType([PropTypes.string, PropTypes.number]).isRequired,
```
ajkl2533, line 30: I would probably extract this same as `AccordionItemId`

PR #43, opened 2021-01-29
feat(Datatable): add dumb reference table

ajkl2533:


src/components/Datatable/Table/temp/IndeterminateCheckbox.tsx
```diff
@@ -0,0 +1,43 @@
+// TODO: move this as next state of the `forms/Checkbox`
+
+import React, { forwardRef, useEffect, useRef } from 'react';
+import PropTypes from 'prop-types';
+
+const useCombinedRefs = (...refs): React.MutableRefObject<HTMLInputElement> => {
```
PrincAm, line 6: It is a bit tricky part. Please why do we need to combine those references?

PrincAm (approved):
Great kick-off! Basic functionality is there and working, styles will be updated within next tasks. I just didn't get the handling of IndeterminateCheckbox's references, maybe it would deserve a comment.

src/components/Datatable/Table/temp/IndeterminateCheckbox.tsx
ajkl2533: Don't worry about that, I'm not sure if we need that at all, it copied from some example

PrincAm: Okey then

ajkl2533: @PrincAm provided a link with explanation

PR #49, opened 2021-02-01
feat(Filters): add fields handling logic

PrincAm:
Please, check only last 8 commits starting by message "feat(BottomBar): add BottomBar component".

What is missing
- Input components
- Bottom bar styling
- Storybook


src/theme/icons/plus.ts
```diff
@@ -0,0 +1,20 @@
...
+ */
+import {
+  IconDefinition,
+  IconName,
+  IconPrefix,
+} from '@fortawesome/fontawesome-svg-core';
+
+export const width = 448;
+export const height = 512;
+export const unicode = 'e010';
```
ajkl2533, line 13: :) it's hexadecimal, so next number after `e009` is not `e010` but `e00a`. I got also caught  :D

src/theme/icons/index.ts
```diff
@@ -13,3 +13,4 @@ export { chevronDown } from './chevronDown';
 export { arrowUp } from './arrowUp';
 export { quoteSquare } from './quoteSquare';
 export { caretDown } from './caretDown';
+export { plus } from './plus';
```
ajkl2533, line 16: Don't forget to add to the `Icon.enums.ts` file. I'll refactor this in the TableHeadCell PR and move that enum into `icons` folder

src/theme/icons/plus.ts
PrincAm: oh man, good catch! I even didn't stop to think about it

src/components/Filters/FilterRow/FilterRow.tsx
```diff
@@ -30,42 +32,114 @@ const SplitField = styled.div<SplitFieldProps>`
   }
 `;
 
+const getDefaultCondition = (dataPoints, dataPointValue) =>
+  dataPoints
```
ajkl2533, line 36: let's create memoized `getDatapointConditions` since it is used here and also in next `getConditionInput`
```js
const getDatapointConditions = R.memoizeWith((dataPoints, dataPointValue) => R.pipe(
  R.find(R.propEq('value', dataPointValue)),
  R.prop('conditions')
)(dataPoints)
```

src/components/Filters/FilterRow/FilterRow.tsx
```diff
@@ -1,21 +1,23 @@
 import React from 'react';
```
ajkl2533, line 1: [Not necessary] This file could get some Ramda-love :D

src/components/Filters/FilterRow/FilterRow.tsx
PrincAm: Perfect, I've been a bit worry with the code optimization, a lot of listings. There are also other adepts for Ramda, recommendations are appreciated

src/components/Filters/FilterRow/FilterRow.tsx
PrincAm: Understand, will have a look 🙂

src/components/Filters/hooks/useFilterRow.ts
```diff
@@ -0,0 +1,59 @@
+import { useEffect, useState } from 'react';
+
+import { DataPoint } from '../Filters.types';
+import { useFilterRowType } from './useFilterRow.types';
+import { Option } from '../Select/Select.types';
+
+export const normalizeOptions = ({ value, label }: Option): Option => ({
```
ajkl2533, line 7: so this should be probably:
```js
export const normalizeOptions = ({ value, label }: Condition): Option =>
```

ajkl2533: Oh I see so this should be actually:
```js
export const normalizeOptions = <O extends Option>({ value, label }: O): Option => {...}
```
and then used as
```js
dataPoint: normalizeOptions<DataPoint>(dataPoint),
condition: normalizeOptions<Condition>(condition),

File: src/components/Filters/Filters.tsx [hunk 6]
```diff
@@ -1,30 +1,181 @@
...
+    const newRow = {
+      operator: newFilters[0].operator,
+      dataPoint: dataPoints[0].value,
+      condition: getDefaultCondition(dataPoints),
+      input: '',
+      isApplied: false,
+    };
+    newFilters.push(newRow);
+
+    setFiltersValues(newFilters);
```
ajkl2533, line 112: I'd rather do 
```js
setFiltersValues([...newFilters, newRow]);
```
instead of `.push`

src/components/Filters/hooks/useFilterRow.ts
ajkl2533, line 7: This should be 
```js
export const normalizeOptions = <O extends Option>({ value, label }: O): Option => ({
```
and then used as
```js
dataPoint: normalizeOptions<DataPoint>(dataPoint),
condition: normalizeOptions<Condition>(condition),
```

src/components/Filters/FilterRow/FilterRow.tsx
ajkl2533: There are also some other candidates like assigning values in hadlers in `Filters.tsx` but at least you don't mutate original values outside those methods. But I assume we can Ramdify later on

src/components/Filters/Filters.tsx
PrincAm: Yes, agree. Changed

src/components/Filters/hooks/useFilterRow.ts
PrincAm: Nice! Thanks for the code examples

src/components/Filters/FilterRow/FilterRow.tsx
PrincAm: I've covered a few functions by Ramda here, `Filters.tsx` will leave for next PR

ajkl2533 (approved):

PR #50, opened 2021-02-01
feat(TableHeadCell): style table head cells

ajkl2533:
Closes FEP-1562


src/components/Datatable/Table/Table.tsx
```diff
@@ -49,15 +49,15 @@ const StyledTableBody = styled.thead<{ isTableSticky: boolean }>`
     `};
 `;
 
-const Table: React.FC<TableProps<Data>> = ({
+const Table = <D extends Record<string, unknown>>({
```
PrincAm, line 52: Please, what is the advantage of using Record type here? If I will add new prop then it still fails, right?

src/components/Datatable/Table/TableHeadCell/SortingIcon.tsx
```diff
@@ -0,0 +1,41 @@
...
+    $isActive &&
+    css`
+      font-size: ${pxToRem(14)};
+    `};
+`;
+
+const SortingIcon: React.FC<SortingIconProps> = React.memo(
+  ({ isSorted, isSortedDesc }) => {
+    if (isSortedDesc)
+      return <StyledIcon name={SSCIconNames.sortDown} $isActive />;
```
PrincAm, line 28: Does have `$` in prop name special meaning?

PrincAm (approved):
Looks good! There is some commented code which can be removed later on. Just a question about the typing for my better understanding to use same pattern.

src/components/Datatable/Table/Table.tsx
ajkl2533: `D` is generic type and we need to add some constraints to it because it would be too wide for typescript. This way we tell TS that the D will have some shape of JS object where keys will be strings and we don't care about values type. 
This will give us a way how to pass type which describes the table data type which will be different for each table.
We can use it then like:
```js
type Data = {
  id: number;
  ipAddress: string;
  ...
}

...

return <Table<Data> ... />
```
And the `Table` component will know type will have the data we pass in. 

The `Record<string, unknown>` notation we use instead of `{}` because `{}` means any non-nullish value but Record<> means any object.

src/components/Datatable/Table/TableHeadCell/SortingIcon.tsx
ajkl2533: A common question these times :D Yep that is the notation for transient props for styled-component (https://styled-components.com/docs/api#transient-props). That means those props are not passed to DOM

src/components/Datatable/Table/Table.tsx
PrincAm: Thanks for clear explanation! 🙏

src/components/Datatable/Table/TableHeadCell/SortingIcon.tsx
PrincAm: Oh yes, now I remember, new transient props. Nice!

youssefssc (approved):
LGTM, one small comment: in the design the border top is transparent/same as cell color

ajkl2533:
@youssefssc checked with Mate top border is correct

PR #51, opened 2021-02-02
Add tab sizes and new styles to use in footprint, keep backwards compatibility

jurajhrib:
Figma link: https://www.figma.com/file/vnzYfrwIISRYkr2iYE5TcD/Platform-%3A-Atoms?node-id=314%3A0

This change is backwards compatible - any use of Tabs will look and function the same, we have just added new sizes and consistency between them. Also some props to customize spacings have been added, small fix of color leaking into DOM.

This component has been created in order to implement new time range control in the Digital Footprint section.


src/components/Tabs/Tab.tsx
```diff
@@ -50,9 +24,10 @@ const Tab: React.FC<TabProps> = ({
 
   return (
     <Label
+      $color={color}
       $isSelected={isSelected}
       as={isLink ? RouterLink : 'a'}
-      color={color}
+      size={size}
```
vaclav-zeman, line 30: any reason you made color transient but not size?

src/components/Tabs/Tabs.stories.tsx
```diff
@@ -82,3 +82,33 @@ export const RoutableTabs: Story = () => {
     </MemoryRouter>
   );
 };
+
+export const TabSizes: Story = () => (
+  <>
+    <Tabs selectedValue="one" size="h2" onSelectTab={action('Select Tab')}>
```
vaclav-zeman, line 88: Do we have a story with no size defined? Just so visual regression covers the default case

vaclav-zeman (approved):
Looks really good, gj! 🎉

src/components/Tabs/Tabs.stories.tsx
jurajhrib: @vaclav-zeman good point, will add. I guess we could even split every size into separate stories. Maybe Radek will have some ideas too.

src/components/Tabs/Tab.tsx
jurajhrib: @vaclav-zeman size does not spill into DOM, but color gets passed into styled

src/components/Tabs/Tabs.stories.tsx
jurajhrib: @vaclav-zeman actually, I've just realized that the first story "Only Tabs" is without size (defaults to h3). Is it okay like that?

vaclav-zeman: yep, for sure 👍

src/components/Tabs/TabLabel.tsx
```diff
@@ -0,0 +1,99 @@
...
+  [TabSizes.sm]: smallUnderlineSize,
+};
+
+const textSizes = {
+  [TabSizes.lg]: largeTextSize,
+  [TabSizes.md]: mediumTextSize,
+};
+
+const underlineTab = css<LabelProps & { size: Sizes; variant: Variants }>`
+  ${({ size }) => underlineSizes[size] || underlineSizes[TabSizes.md]};
```
ajkl2533, line 49: what about?
```suggestion
  ${({ size = TabSizes.md }) => underlineSizes[size]};
```

ajkl2533: Or you can remove the default entirely since this component will be used only in Tabs component and default size is set up there.

ajkl2533 (approved):
LGTM, good job 👍

src/components/Tabs/TabLabel.tsx
```diff
@@ -0,0 +1,99 @@
...
+  &:visited {
+    color: ${getColor('graphite4B')};
+  }
+`;
+
+const textTab = css<LabelProps & { size: Sizes; variant: Variants }>`
+  ${({ size }) => textSizes[size] || textSizes[TabSizes.md]};
+
+  color: ${({ $isSelected, $color }) =>
+    $isSelected ? getColor($color) : getColor('graphiteHB')};
```
vaclav-zeman, line 72: Noticed here when I try to use it that this is incorrect, should be:

default: `blueberryClassic` (I'd set this as default value of $color), selected: `graphite4B`, hover/focus: `lighten($color, xx)`

src/components/Tabs/TabLabel.tsx
vaclav-zeman, line 49: The default seems way too big, maybe TabSizes.sm?
image

PR #53, opened 2021-02-03
feat(DatePicker): add component

PrincAm:
Unfortunately for custom styling of DateRangePicker is needed to overwrite its style classes https://github.com/airbnb/react-dates#overriding-styles

Moment is required dependency

What is missing
- Multiselect component?
- SingleDatePicker?
- Move input components into separate folder
- Bottom bar styling
- Storybook


youssefssc (approved):
LGTM with a couple of comments/questions. 
When typing in the dates in MM/DD/YYYY format, I see the `input` object changes correctly but no error is shown when the start date is after the end date. 
Is it possible to have the help text show below the date picker to avoid having a popup inside a popup?

PrincAm:
@youssefssc Good point. I've discussed the dates issue with UX and we chose option to make input fields read only. It means manual input is disabled. Regarding the help modal, UX is familiar with its behavior. It is not ideal but displaying another modal underneath is not better, IMO.

youssefssc:
@PrincAm Making input read only works great! 
I didn't mean another modal underneath for help but to extend the date modal vertically to include the help content. I will attach an illustration.

PrincAm:
@youssefssc The help modal is native behavior of react-dates and works. Make it different would require styles overwriting etc and I am glad I don't have to mess a lot with modal itself because of position calculating. Even more, the help is so minor that I wouldn't spend more time than necessarily. How I wrote, UX is aware and fine with modal inside modal.

package.json
```diff
@@ -42,11 +42,14 @@
     "@types/react": "^16.9.38",
     "@types/styled-components": "^5.1.0",
     "csstype": "^3.0.0-beta.4",
+    "moment": "^2.29.1",
```
ajkl2533, line 45: Damn, I decided to go with `dayjs` package which provides the same API as `moment` but is way smaller. Is the usage of `moment` hard requirement of the `react-dates`? Or can we replace it?

ajkl2533 (approved):
All good except we need to agree on using `Moment`/`dayjs` library

PR #57, opened 2021-02-05
Add Tabs accessibility

jurajhrib:
Mainy to prevent bootstrap reboot to mess with our styles.


ajkl2533 (approved):

vaclav-zeman:
As discussed on slack, can we rather use a `button` component as a default (instead of `a` without href) so it's accessible without any additional code?

PR #58, opened 2021-02-05
feat(Datatable): add ControlModule

youssefssc:
closes FEP-1523 and FEP-1608


src/components/Datatable/ControlModule/ControlModule.tsx
```diff
@@ -0,0 +1,112 @@
...
+import { FlexContainer } from '../../FlexContainer';
+import { BatchModule } from '../BatchModule';
+import { SearchBar } from '../../forms/SearchBar';
+import { ToolsTabItem } from '../ToolsTabItem';
+import { ControlModuleProps } from './ControlModule.types';
+import { Filters } from '../../Filters';
+import { renderSuggestionFilter } from '../../forms/SearchBar/SearchSuggestionFormats';
+
+const ControlModuleContainer = styled(FlexContainer)`
+  background-color: ${getColor('graphite3H')};
```
ajkl2533, line 17: no need, it is already done in Datatable component

src/components/Datatable/ControlModule/ControlModule.tsx
```diff
@@ -0,0 +1,112 @@
...
+`;
+
+const FiltersContainer = styled(FlexContainer)`
+  background-color: ${getColor('graphite5H')};
+`;
+
+const ControlModule: React.FC<ControlModuleProps> = ({
+  toolsTabItems = [],
```
ajkl2533, line 25: Those items are strictly given by design (Show/Hide, Filter, Groups, View), we will not provide API for custom controls. We just decide by given datatable configuration which tools we should show

src/components/Datatable/ControlModule/ControlModule.tsx
```diff
@@ -0,0 +1,112 @@
...
+  onSearch,
+  renderSearchSuggestion = renderSuggestionFilter,
+  data: filtersData,
+  dataPoints: filtersDataPoints,
+  onApply: onApplyFilter,
+  onCancel: onCancelFilter,
+  onChangeFilter,
+  onClose,
+  actions,
+  isFilterOpen = false,
```
ajkl2533, line 36: please rename to `defaultIsFilterOpen`

src/components/Datatable/ControlModule/ControlModule.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+const FiltersContainer = styled(FlexContainer)`
+  background-color: ${getColor('graphite5H')};
+`;
+
+const ControlModule: React.FC<ControlModuleProps> = ({
+  toolsTabItems = [],
+  placeholder: searchPlaceholder = 'Search',
+  onSearch,
+  renderSearchSuggestion = renderSuggestionFilter,
```
ajkl2533, line 28: This shouldn't be also customizable we are deciding how we render suggestions in datatable.

src/components/Datatable/ControlModule/ControlModule.tsx
```diff
@@ -0,0 +1,112 @@
...
+          },
+        }),
+      };
+    });
+    setTools(updatedTools);
+  }, [toolsTabItems]);
+
+  return (
+    <ControlModuleContainer flexDirection="column">
+      <FlexContainer padding={{ vertical: 0.5 }}>
```
ajkl2533, line 67: I would re-do this whole section like:
```js
<FlexContainer alignItems="center" padding={{vertical: 0.4, horizontal: 0.8}}>
  <FlexContainer flexShrink={1} margin={{ right: 1.6 }}>Tools</FlexContainer>
  <FlexContainer flexGrow={1}>SearchBar</FlexContainer>
</FlexContainer>

File: src/components/Datatable/ControlModule/ControlModule.tsx [hunk 6]
```diff
@@ -0,0 +1,112 @@
...
+        <Col>
+          <SearchBar
+            placeholder={searchPlaceholder}
+            renderSearchSuggestion={renderSearchSuggestion}
+            onSearch={onSearch}
+          />
+        </Col>
+      </FlexContainer>
+      {areFiltersOpen && (
+        <FiltersContainer padding={1}>
```
ajkl2533, line 85: This padding should be `{top:1.2, bottom:0.8, horizontal: 0.8}` by design

src/components/Datatable/ControlModule/ControlModule.tsx
```diff
@@ -0,0 +1,112 @@
...
+            dataPoints={filtersDataPoints}
+            onApply={onApplyFilter}
+            onCancel={onCancelFilter}
+            onChangeFilter={onChangeFilter}
+            onClose={onCloseFilter}
+          />
+        </FiltersContainer>
+      )}
+
+      <BatchModule actions={actions} />
```
ajkl2533, line 97: Why did you put BatchModule here? It should be outside

src/components/Datatable/ControlModule/ControlModule.tsx
```diff
@@ -0,0 +1,112 @@
...
+  onCancel: onCancelFilter,
+  onChangeFilter,
+  onClose,
+  actions,
+  isFilterOpen = false,
+}) => {
+  const [areFiltersOpen, setAreFiltersOpen] = useState(isFilterOpen);
+  const [tools, setTools] = useState(toolsTabItems);
+
+  const onCloseFilter = () => {
```
PrincAm, line 41: Please, use `handle` prefix for event handler. `handleCloseFilter`

src/components/Datatable/BatchModule/ElementCounter/ElementCounter.tsx
```diff
@@ -58,7 +58,11 @@ const ElementCounter: React.FC<ElementCounterProps> = ({
   );
 
   return (
-    <H3 margin="none" variant={HeadingVariants.secondary}>
+    <H3
+      margin="none"
+      padding={{ horizontal: 0.4 }}
```
ajkl2533, line 63: this should be done in https://github.com/securityscorecard/design-system/blob/datatable/src/components/Datatable/BatchModule/BatchModule.tsx#L20

youssefssc: Right. moving it there

src/components/Datatable/ControlModule/ControlModule.tsx
```diff
@@ -0,0 +1,112 @@
...
+import { SearchBar } from '../../forms/SearchBar';
+import { ToolsTabItem } from '../ToolsTabItem';
+import { ControlModuleProps } from './ControlModule.types';
+import { Filters } from '../../Filters';
+import { renderSuggestionFilter } from '../../forms/SearchBar/SearchSuggestionFormats';
+
+const ControlModuleContainer = styled(FlexContainer)`
+  background-color: ${getColor('graphite3H')};
+`;
+
```
ajkl2533: Suggestion:
1) let's create ToolBar component and encapsulate the toolTabsItems in instance
2) remove this toolTabsItems completely
3) define toolTabsItems in map in the ToolBar component (do not pass anything from outside exept some custom handlers for activate/deactivate), like:
```js
const tools = [
  {
    label: '',
    icon: ''
  },
  ...
]
```
4) move `useEffect` into the ToolBar component
5) for show/hide filters panel we should use DatatableContext and should be controled in the ToolBar component

src/components/Filters/Filters.tsx
```diff
@@ -51,8 +51,8 @@ const Filters: React.FC<FiltersProps> = ({
   }, [filtersValues, dataPoints]);
 
   const callOnChange = () => {
-    if (typeof onChange === 'function') {
-      onChange(filtersValues);
+    if (typeof onChangeFilter === 'function') {
+      onChangeFilter(filtersValues);
```
ajkl2533, line 55: I don't see any reason to change the name of this property. It's not called only on filters change but on any change within Filters component so `onChange` is completely fine.

src/components/Datatable/ControlModule/ControlModule.tsx
ajkl2533: 6. for the Toolbar I would add `state` property which defines initial state, e.g. filters are opened, initial show/hide state...

src/components/Filters/Filters.tsx
youssefssc: It created a type conflict because there is onChange prop for SearchBar / input

src/components/Datatable/ControlModule/ControlModule.tsx
youssefssc: I am confused by 3. Why do you think the tools icons/labels is better to be hardcoded and not flexible for future tables with different tools?

src/components/Filters/Filters.tsx
ajkl2533: I'm confused now, how prop name in Filters component can clash with a prop in SearchBar component?

youssefssc: It is confusing :) Because they are merged in
 `interface ControlModuleProps extends SearchBarProps, FiltersProps { ... }`

ajkl2533: ah ok, so in that case it would be probably better not extending but recreating them 
```
interface ControlModuleProps {
  onFilterChange: () => {},
  onSearchChange: () => {},
  ...
}
because it's problem of the ControlModule not Filters that there are clashing prop names

File: src/components/Datatable/ControlModule/ControlModule.tsx [hunk 12]
```diff
@@ -0,0 +1,133 @@
...
+  hasColumnVisibility = true,
+  // defaultHiddenColumns = [],
+  hasColumnOrdering = true,
+  // defaultColumnOrder,
+  hasCustomViews = true,
+  // defaultCustomView,
+  columnVisibilityActions,
+  filteringActions,
+  columnOrderingActions,
+  customViewsActions,
```
ajkl2533, line 33: What are those 4 props?

src/components/Datatable/ControlModule/ControlModule.types.ts
```diff
@@ -0,0 +1,5 @@
+import { ControlsConfig } from '../Datatable.types';
+
+export interface ControlModuleProps<D extends ControlsConfig<D>> {
+  someprop?: boolean;
```
ajkl2533, line 4: What is this?

src/components/Datatable/ControlModule/ControlModule.tsx
```diff
@@ -0,0 +1,133 @@
...
+          )}
+          {hasFiltering && (
+            <ToolsTabItem
+              iconName={SSCIconNames.filter}
+              label="Filter"
+              toolState={defaultIsFilteringOpen ? 'active' : 'default'}
+              {...filterToolActions}
+            />
+          )}
+          {hasColumnOrdering && (
```
ajkl2533, line 75: Grouping and Column ordering are two different features. Grouping API wasn't yet even proposed.
Column ordering is part of the `Show/Hide` feature.

src/components/Datatable/ControlModule/ControlModule.tsx
```diff
@@ -0,0 +1,133 @@
...
+            <ToolsTabItem
+              iconName={SSCIconNames.sitemap}
+              label="Group"
+              {...columnOrderingActions}
+            />
+          )}
+          {hasCustomViews && (
```
ajkl2533, line 82: I would completely remove Views and Group from the UI and from API. Those are nice-to-have features and we don't know yet how the API should look like and even if we will be able to implement those in Q1.

src/components/Datatable/Datatable.tsx
```diff
@@ -8,13 +8,15 @@ import { getBorderRadius, getColor } from '../../utils/helpers';
...
-import { BatchModule } from './BatchModule';
 import DatatableContext from './DatatableContext';
 import {
   ControlsConfig,
   DatatableProps,
   ExtendedTableConfig,
 } from './Datatable.types';
+import { ControlModule } from './ControlModule';
+import { BatchModule } from './BatchModule';
+import { mockControlModuleProps } from './mocks/controlModule';
```
ajkl2533, line 19: Since Datatable is already released there should not be any mocks in the components.

src/components/Datatable/ToolsTabItem/ToolsTabItem.tsx
```diff
@@ -144,21 +145,6 @@ const ToolsTabItem: React.FC<ToolsTabItemProps> = ({
...
-    PropTypes.oneOf<Types>(Object.values(IconTypes)),
-    PropTypes.string,
-  ]),
-  margin: SpacingSizeValuePropType,
-  isDisabled: PropTypes.bool,
-  className: PropTypes.string,
-  onToolActivate: PropTypes.func,
-  onToolDeactivate: PropTypes.func,
-};
+ToolsTabItem.propTypes = { ...ToolsTabItemPropType };
```
ajkl2533, line 148:
```suggestion
ToolsTabItem.propTypes = ToolsTabItemPropType;
```

src/components/Datatable/mocks/toolsTabItems.ts
```diff
@@ -0,0 +1,58 @@
+import { ToolsTabItem } from '../ToolsTabItem/ToolsTabItem.types';
```
ajkl2533, line 1: I think we can remove this file

src/components/Datatable/mocks/controlModule.ts
```diff
@@ -0,0 +1,45 @@
...
+    onCancel: action('onCancel'),
+    onChange: action('onChange'),
+    onClose: action('onClose'),
+  },
+  defaultIsFilteringOpen: false,
+  hasColumnVisibility: true,
+  defaultHiddenColumns: [],
+  hasColumnOrdering: true,
+  defaultColumnOrder: [],
+  toolsTabItems: mockToolsTabItemsFiltersActive,
```
ajkl2533, line 28: I believe this doesn't have to be in the mock

src/components/Datatable/ControlModule/ControlModule.stories.tsx
```diff
@@ -0,0 +1,92 @@
...
+import { mockControlModuleProps } from '../mocks/controlModule';
+
+export default {
+  title: 'components/Datatable/components/ControlModule',
+  component: ControlModule,
+  decorators: [(storyFn) => <MemoryRouter>{storyFn()}</MemoryRouter>],
+  subcomponents: { BatchModule, ToolsTabItem, SearchBar },
+} as Meta;
+
+export const playground: Story<ControlModuleProps<D>> = (args) => (
```
ajkl2533, line 20: Instead of D use either `Data` from  https://github.com/securityscorecard/design-system/blob/alpha/src/components/Datatable/Datatable.stories.tsx#L35 or `Record<'string', unknown>`

src/components/Datatable/ControlModule/ControlModule.tsx
```diff
@@ -0,0 +1,133 @@
...
+import { SSCIconNames } from '../../../theme/icons/icons.enums';
+import { ControlModuleProps } from './ControlModule.types';
+import { Filters } from '../../Filters';
+import { renderSuggestionFilter } from '../../forms/SearchBar/SearchSuggestionFormats';
+
+const FiltersContainer = styled(FlexContainer)`
+  background-color: ${getColor('graphite5H')};
+`;
+
+const ControlModule: React.FC<ControlModuleProps> = ({
```
ajkl2533, line 18: in order to fix the TS error in here, you have to change the ControlModule signature in similar fashion like https://github.com/securityscorecard/design-system/blob/alpha/src/components/Datatable/Datatable.tsx#L47-L57

So it should be:
```js
const ControlModule = <D extends Record<string, unknown>>({
  hasSearch = true,
  searchConfig,
  hasFiltering = true,
  filtersConfig,
  defaultIsFilteringOpen = true,
  hasColumnVisibility = true,
  // defaultHiddenColumns = [],
  hasColumnOrdering = true,
  // defaultColumnOrder,
  hasCustomViews = true,
  // defaultCustomView,
  columnVisibilityActions,
  filteringActions,
  columnOrderingActions,
  customViewsActions,
}: ControlModuleProps <D>): React.ReactElement => {
```
And then in `Datatable.tsx` you will use it like
```js
{controlsConfig.isControlsEnabled && (
         <ControlModule<D>
            {...defaultControlsConfig}
          />
        )}         
```

src/components/Datatable/Datatable.tsx
```diff
@@ -83,8 +89,12 @@ const Datatable = <D extends Record<string, unknown>>({
       }}
     >
       <StyledDatatable flexDirection="column">
-        <FlexContainer>ToolbarModule</FlexContainer>
-        <FlexContainer>Filters</FlexContainer>
+        {controlsConfig.isControlsEnabled && (
+          <ControlModule
+            {...defaultControlsConfig}
```
ajkl2533, line 94: There is merged config on line 69 to 72, you should be using that. This will use just the default properties and it cannot be affected from outside.

src/components/Datatable/ControlModule/ControlModule.tsx
ajkl2533: disregard that :)

src/components/Datatable/ControlModule/ControlModule.types.ts
youssefssc: removed, it was just a placeholder to test fixing type errors

src/components/Datatable/ControlModule/ControlModule.tsx
youssefssc: Thanks! was confused by the split, now added `hasGrouping`

src/components/Datatable/ControlModule/ControlModule.tsx
youssefssc: set to false in defaultConfig

src/components/Datatable/Datatable.tsx
youssefssc: mocks removed

src/components/Datatable/mocks/controlModule.ts
youssefssc: removed

src/components/Datatable/Datatable.tsx
youssefssc: updated to use custom controls config

src/components/Datatable/ControlModule/ControlModule.tsx
youssefssc: Thanks! I used this type and it fixed most of the conflicts. I still have one type error left in ControlModule storybook though which I couldn't remove

src/components/Datatable/Datatable.tsx
ajkl2533: Now it uses just the config passed as a property. There is this line
```js
const {
    defaultHiddenColumns,
    defaultColumnOrder,
  }: ControlsConfig<D> = mergeRight(defaultControlsConfig, controlsConfig);
```
sou you should add `...restControlsConfig` in the destructure and then do 
```js
<ControlModule
  defaultHiddenColumns={defaultHiddenColumns}
  defaultColumnOrder={defaultColumnOrder}
  {...restControlsConfig}
/>

src/components/Datatable/ControlModule/ControlModule.tsx
ajkl2533: I was looking into that and it seems there is some kind of issue with validation between prop-types and TypeScript. 
Quick hacky fix, for now, is to remove the `as Meta` in default export in storybook file.

ajkl2533 (approved):
LGTM

PR #62, opened 2021-02-09
feat(DateRangePicker): add component

PrincAm:
Only DataRangePicker.tsx, styles, prop types and dependencies were updated compared to https://github.com/securityscorecard/design-system/pull/53


src/components/Filters/DateRangePicker/DateRangePicker.tsx
```diff
@@ -0,0 +1,64 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import DatePicker from 'react-datepicker';
+import 'react-datepicker/dist/react-datepicker.css';
```
ajkl2533, line 5: 🤦  😅

src/components/Filters/FilterRow/FilterRow.tsx
```diff
@@ -123,7 +124,11 @@ const FilterRow: React.FC<FilterRowProps> = ({
   };
 
   const handleInputChange = (event) => {
-    onInputChange(event.target.value, index);
+    if (event.target) {
```
ajkl2533, line 127: Why do we need to split this in condition?

ajkl2533 (approved):
I don't see anything wrong :)

src/components/Filters/FilterRow/FilterRow.tsx
PrincAm: It is generic handler
- if input component is `Input` then event is Event type
- if input component is DateRangePicker then event is object `{startDate, endDate}`
- if input component is Select then event is object `{value, label}` - not yet implemented
- if input component is SingleDatePicker then event is date - not yet implemented

Maybe event is misleading name now, I can change it. Also the condition will change when I will add other components.

src/components/Filters/DateRangePicker/DateRangePicker.tsx
PrincAm: Hm, what can we do. Now I believe there is not better option than `react-datepicker` 😄

src/components/Filters/FilterRow/FilterRow.tsx
ajkl2533: ah ok I was misled by the name

PrincAm: Ok, I will rename it in next PR

PR #67, opened 2021-02-10
feat(Number): add component

PrincAm:
Only numbers restriction I took from https://stackoverflow.com/questions/43687964/only-numbers-input-number-in-react It also works for copy pasting.

I've also added placeholder to Input and unified placeholders color.


src/components/Filters/mocks/options.ts
```diff
@@ -55,7 +55,7 @@ export const dataMock: Filter[] = [
     operator: Operators.and,
     dataPoint: 'lemon',
     condition: 'strawberry',
-    input: 'fruity',
+
```
youssefssc, line 58: empty line

youssefssc (approved):
LGTM

src/components/Filters/inputs/Number/Number.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+const Number: React.FC<NumberProps> = ({ value = '', onChange }) => {
+  const handleInputChange = (event) => {
+    const newValue = event.target.validity.valid ? event.target.value : value;
+    onChange(newValue);
+  };
+
+  return (
+    <StyledInput
+      pattern="[0-9]*"
```
ajkl2533, line 15: hm, why? :D when you pass `type="number"` users cannot put anything else than number in field

ajkl2533 (approved):
LGTM, just change input type to `number` and remove `pattern` attribute

src/components/Filters/inputs/Number/Number.tsx
PrincAm: `type="number"` was my first choice but then I tried also other browsers and was pretty disappointed
- Chrome - user can insert only numbers
- Firefox - user can insert other characters but then will get error (red input border)
- Safari - user can insert other characters and will not get error
I assume it is more suitable for form submit, not alone. Even more, design is without increment arrow buttons. So I ended up with this solution.

ajkl2533: so can we do `type="number"` and `pattern="[0-9]*"`? because if we don't use `number` type we can't use other attributes like `min,`max`,`step`
I don't care about the increment arrow, it's browser-specific so we probably shouldn't override it and also it's solvable through css if we really need it

PrincAm: @ajkl2533 updated to `type="number"` and overall simplified. The extra event handler is gone, not needed

PR #68, opened 2021-02-11
feat(Select): support Select component

PrincAm:


src/components/Filters/FilterRow/FilterRow.tsx
```diff
@@ -78,6 +80,44 @@ const getOperatorOptions = curry((operatorValue) =>
...
 const getDataPointOptions = map(normalizeOptions);
 
+const renderComponent = (Component, value, onChange) => {
+  if (!Component) {
+    return null;
+  }
+
+  /* Select */
+  if (
+    typeof Component === 'object' &&
```
ajkl2533, line 90: https://char0n.github.io/ramda-adjunct/2.28.0/RA.html#.isObj

src/components/Filters/FilterRow/FilterRow.tsx
```diff
@@ -126,6 +166,8 @@ const FilterRow: React.FC<FilterRowProps> = ({
   const handleInputChange = (value) => {
     if (value.target) {
       onInputChange(value.target.value, index);
+    } else if (value.value) {
```
ajkl2533, line 169: I would rather hasPath/hasProp here since if `value.value` will be `0` or empty string it's still false, same goes for `value.target`

src/components/Filters/FilterRow/FilterRow.tsx
```diff
@@ -78,6 +80,44 @@ const getOperatorOptions = curry((operatorValue) =>
 
 const getDataPointOptions = map(normalizeOptions);
 
+const renderComponent = (Component, value, onChange) => {
+  if (!Component) {
```
ajkl2533, line 84: maybe would be good to check if the `Component` is present and is valid React Element via `React.isValidElement`
```js
if (isUndefined(Component && !React.isValidElement(Component))  return null;

File: src/components/Filters/FilterRow/FilterRow.tsx (hunk 2 above)
Comment (PrincAm) [reply src/components/Filters/FilterRow/FilterRow.tsx:169]: Yep, good point with 0 and empty string. Changed

File: src/components/Filters/FilterRow/FilterRow.tsx (hunk 3 above)
Comment (PrincAm) [reply src/components/Filters/FilterRow/FilterRow.tsx]: Good tip with `React.isValidElement`, updated

Comment (ajkl2533) [reply src/components/Filters/FilterRow/FilterRow.tsx]: the condition looks strange (I wrote it wrongly and you copy/pasted) 😆 
```js
if (isUndefined(Component) || !React.isValidElement(Component)) return null
```

ajkl2533 (dismissed):

src/components/Filters/FilterRow/FilterRow.tsx
PrincAm: @ajkl2533 `Component` is here wrapped into argumentless function therefore `React.isValidElement` doesn't apply. I need to wrap the functional Component when using `useState` in `useFilterRow` hook, please see my last commit, otherwise Component is invoked and fails. A bit strange error. At least a believe it is the explanation https://dev.to/sarioglu/how-to-store-a-function-using-react-usestate-4ffh

PrincAm: I simplified the hook and omitted storing of component itself in local state, still `isValidElement` checks reference to component as invalid
```
React.isValidElement(Component) -> false
React.isValidElement(<Component />) -> true
```

src/components/Filters/hooks/useFilterRow.ts
```diff
@@ -37,27 +37,23 @@ export const useFilterRow = (
...
     setDataPoint(dataPointOptions);
     setConditions(conditionsOptions);
     setCondition(conditionOptions);
-    // Wrap Input into argumentless function to not be invoked immediately
-    setInputComponent(() => Input);
   }, [dataPoints, dataPointValue, conditionValue]);
 
   return {
-    dataPoint,
+    dataPoint: dataPoint && normalizeOptions<DataPoint>(dataPoint),
```
ajkl2533, line 54: please use explicit check for `dataPoint` and `condition`

ajkl2533 (approved):

PR #76, opened 2021-02-12
refactor(Filters): rename props, fns, folder

PrincAm:
Renaming, no code logic change.


src/components/Filters/Filters.tsx
```diff
@@ -6,22 +6,22 @@ import styled from 'styled-components';
...
 
-const generateId = ({ operator, dataPoint, condition }, index) =>
-  `${operator}-${dataPoint}-${condition}-${index}`;
+const generateId = ({ operator, field, condition }, index) =>
+  `${operator}-${field}-${condition}-${index}`;
 
-const getDefaultCondition = (dataPoints) =>
-  dataPoints[0].conditions.find(({ isDefault }) => isDefault).value;
+const getDefaultCondition = (fields) =>
+  fields[0].conditions.find(({ isDefault }) => isDefault).value;
```
ajkl2533, line 17: I'm thinking: we should make default the first condition in array if no `isDefault` is defined

PrincAm: Definitely, it will break now. Let me fix that in next PR

ajkl2533 (approved):
LGTM

PR #78, opened 2021-02-15
fix(Cell): add check for onClick handler

ajkl2533:


src/components/Datatable/Table/TableCell/renderers/Text.tsx
```diff
@@ -38,12 +38,18 @@ const Text: React.FC<LinkCellRendererProps<string | number>> = ({
     return <TableCellText>{returnValue}</TableCellText>;
   }
 
+  const handleClick = () => {
```
vaclav-zeman, line 41: Cells prob render a lot, maybe `useCallback`?

vaclav-zeman (approved):
🚢 it

PR #79, opened 2021-02-15
feat(Filters): update storybook and add final tweaks

PrincAm:


src/components/Filters/BottomBar/BottomBar.tsx
```diff
@@ -1,9 +1,32 @@
...
 import { BottomBarProps } from './BottomBar.types';
+import {
+  getColor,
+  getFontSize,
+  getFontWeight,
+  getLineHeight,
+  pxToRem,
+} from '../../../utils/helpers';
+
+const FiltersNotice = styled.div`
```
ajkl2533, line 17: `Paragraph`, size `md`, variant `secondary`, margin `{right: 0.8}`, as `div`

src/components/Filters/BottomBar/BottomBar.tsx
```diff
@@ -1,9 +1,32 @@
...
+  color: ${getColor('graphite2B')};
+  font-size: ${getFontSize('md')};
+  line-height: ${getLineHeight('md')};
+  margin-right: ${pxToRem(16)};
+`;
+
+const LoadingText = styled.span`
```
ajkl2533, line 24: `Text`, size `md`, margin `{left: 0.4}`, only font-weight should stay, since `isBold` is using `bold` not `medium`

ajkl2533 (approved):
LGTM, except two minor comments

src/components/Filters/BottomBar/BottomBar.tsx
PrincAm: @ajkl2533 there is no option for white color, right?

PR #82, opened 2021-02-17
feat(icons): add long arrow icons

ajkl2533:


youssefssc (approved):
LGTM
Needs a rebase because there is a couple of storybook shots conflicts

ajkl2533:
@youssefssc no need to rebase we just need to fix the storybook files since it is using current date instead of mocked one, cc @PrincAm

PR #83, opened 2021-02-17
feat(Table): implement selection exclude logic

ajkl2533:
Closes FEP-1582


src/components/Datatable/BatchModule/ElementCounter/ElementCounter.tsx
```diff
@@ -30,22 +44,49 @@ const getCounterContent = (totalLength, selectedLength = 0) => {
...
+  const computedSelectedLength = hasExclusionLogic
+    ? totalLength - selectedLength
+    : selectedLength;
+
+  const content = getCounterContent(totalLength, computedSelectedLength);
+  const handleSelectClick = useCallback(
+    (state) => {
+      setHasExclusionLogic(state);
+    },
+    [setHasExclusionLogic],
```
PrincAm, line 64: For those useCallbacks, is correct assumption that the setter functions never update and will be used just their first instance?

src/components/Datatable/Datatable.stories.tsx
```diff
@@ -181,18 +180,29 @@ export const Default: Story = () => {
     },
     [setIsLoading, setTableData],
   );
+  const logSelectedRows = useCallback((ids, exclude) => {
+    console.log(
+      `OnSelect called for ids: ${ids} to be ${
+        exclude ? 'excluded' : 'included'
+      }`,
+    );
+  }, []);
```
PrincAm, line 189: Here useCallback shouldn't help without dependency

src/components/Datatable/BatchModule/ElementCounter/ElementCounter.tsx
ajkl2533: yep it's internal DT state which isn't customizable and not exposed

src/components/Datatable/Datatable.stories.tsx
ajkl2533: Yep doesn't make much sense here, it doesn't hurt either in this case, will remove

PrincAm (approved):
LGTM, just a question for better understanding

PR #84, opened 2021-02-17
feat(Table): add Pagination

youssefssc:
https://zitenote.atlassian.net/browse/FEP-1583


src/components/Datatable/Pagination/NavButton.tsx
```diff
@@ -0,0 +1,39 @@
...
+  width: ${pxToRem(48)};
+  height: ${pxToRem(32)};
+  padding: ${pxToRem(8, 16)};
+  margin: ${pxToRem(0, 4)};
+  border-radius: ${pxToRem(3)};
+
+  color: ${({ isDisabled }) =>
+    getColor(isDisabled ? 'graphiteB' : 'graphite2B')};
+  background-color: ${({ isDisabled }) =>
+    getColor(isDisabled ? 'transparent' : 'graphite5H')} !important;
```
PrincAm, line 25: `!import` is needed? trick with `&` doesn't work?

src/components/Datatable/Pagination/NavButton.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+     &:hover {
+      color: ${getColor('graphite3B')} !important;
+      border-color: ${getColor('graphite3B')} !important;
+      //  background-color: ${getColor('graphite3B')} !important;
```
PrincAm, line 30: commented code

src/components/Datatable/Pagination/NavButton.tsx
```diff
@@ -0,0 +1,39 @@
...
+      text-decoration: none;
+    }
+`;
+
+// NavButton.propTypes = {
```
PrincAm, line 35: same here, commented code

src/components/Datatable/Pagination/PageButton.tsx
```diff
@@ -0,0 +1,50 @@
...
+  .attrs(() => ({
+    variant: ButtonVariants.text,
+  }))
+  .withConfig<PageButtonProps>({
+    shouldForwardProp: (prop) => prop !== 'isActive',
+  })`
+  display: inline-block;
+  text-align: center;
+  font-size: ${({ isActive }) =>
+    isActive ? getFontSize('lg') : getFontSize('md')};
```
PrincAm, line 25: minnor, `getFontSize(isActive ? 'lg' : 'md')` is shorter 🙂

src/components/Datatable/Pagination/PageButton.tsx
```diff
@@ -0,0 +1,50 @@
...
+  font-weight: ${({ isActive }) =>
+    isActive ? getFontWeight('bold') : getFontWeight('regular')};
```
PrincAm, line 27: same here

src/components/Datatable/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,206 @@
...
+  margin: ${pxToRem(6, 16)};
+`;
+const PageInput = styled(StyledNumber)`
+  width: ${pxToRem(72)};
+  height: ${pxToRem(32)};
+  padding: ${pxToRem(8, 16)};
+`;
+
+const Pagination: React.FC<PaginationProps> = ({
+  gotoPage,
```
PrincAm, line 43: Could you add `on` prefix to fn props? `onGoToPage, onPreviousPage, onNextPage`

src/components/Datatable/Pagination/Pagination.stories.tsx
```diff
@@ -0,0 +1,59 @@
...
+  title: 'components/Datatable/components/Pagination',
+  component: Pagination,
+} as Meta;
+
+export const Default: Story<PaginationProps> = ({ ...args }) => {
+  const [mockPageIndex, setMockPageIndex] = useState(0);
+  const [mockPageCount, setMockPageCount] = useState(args.pageCount);
+  const [isLoading, setIsLoading] = useState(args.isLoading);
+
+  const gotoPage = (page) => {
```
PrincAm, line 17: Could you add `handle` prefix to handlers? `handleGoToPage, ...` Then it is easier to recognize what is component prop and what is handler itself

src/components/Datatable/Pagination/Pagination.stories.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+import Pagination from './Pagination';
+import { PaginationProps } from './Pagination.types';
+
+export default {
+  title: 'components/Datatable/components/Pagination',
+  component: Pagination,
+} as Meta;
+
+export const Default: Story<PaginationProps> = ({ ...args }) => {
```
PrincAm, line 12: Needs to be copy of `args`? Isn't just passing args down enough? `Story = (args) => {}`

src/components/Datatable/Pagination/NavButton.tsx
```diff
@@ -0,0 +1,39 @@
+import styled from 'styled-components';
+
+import { getColor, pxToRem } from '../../../utils/helpers';
+import { IconButton } from '../../IconButton';
+import { IconButtonProps } from '../../IconButton/IconButton.types';
+
+const NavButton = styled(IconButton)
```
ajkl2533, line 7: It doesn't make sense to inherit `IconButton` component when you replace basically every style of that, just use `.button`. Then you don't need the `!important` we shouldn't be used at all.

src/components/Datatable/Pagination/NavButton.tsx
```diff
@@ -0,0 +1,39 @@
...
+  .attrs((props) => ({
+    ...props,
+  }))
+  .withConfig<IconButtonProps>({
+    shouldForwardProp: (prop) => prop !== 'disabled',
```
ajkl2533, line 12: Why are you disallowing forwarding of `disabled` prop? It's `<button>` native attribute which gives you the ability to actually disable button. So you should remove the `withConfig` and rather do:
```js
.attrs((props) => ({
    ...props,
    disabled: isDisabled
  }))

File: src/components/Datatable/Pagination/Pagination.tsx [hunk 11]
```diff
@@ -0,0 +1,206 @@
...
+    const pageNumber = Number(e.target.value);
+    if (pageNumber <= pageCount && pageNumber > 0) {
+      setIsInputPageInvalid(false);
+      gotoPage(pageNumber - 1);
+    } else {
+      setIsInputPageInvalid(true);
+    }
+  };
+
+  useEffect(() => {
```
PrincAm, line 69: I would at least add comments what is going on here and named the magic numbers, like `MID_SEQUENCE` etc. Also would worth to move the logic to custom hook then is easier to cover it by test. I still have a feeling it could be simplified 🙂

src/components/Datatable/Pagination/NavButton.tsx
```diff
@@ -0,0 +1,39 @@
...
+  })`
+  display: inline-block;
+
+  width: ${pxToRem(48)};
+  height: ${pxToRem(32)};
+  padding: ${pxToRem(8, 16)};
+  margin: ${pxToRem(0, 4)};
+  border-radius: ${pxToRem(3)};
```
ajkl2533, line 20: we don't have 3px `border-radius`. We only use 4px (via `getBorderRadius` helper) or use `50em` (for pills like elements) or 100% (for rounded elements). If the 3px is used in Figma you should write to Mate, as it should be considered design error.

src/components/Datatable/Pagination/PageButton.tsx
```diff
@@ -0,0 +1,50 @@
...
+  line-height: ${getLineHeight('lg')};
+  cursor: pointer;
+
+  color: ${getColor('graphite4B')} !important;
+  background-color: ${({ isActive }) =>
+    getColor(isActive ? 'graphiteH' : 'transparent')};
+
+  padding: ${pxToRem(6, 16)};
+  margin: ${pxToRem(0, 6)};
+  height: ${pxToRem(32)};
```
PrincAm, line 37: Button width is changing based on content and then pagination is jumping. I would set fixed width.

src/components/Datatable/Pagination/NavButton.tsx
youssefssc: removing it, it just wasn't working before

src/components/Datatable/Pagination/Pagination.tsx
ajkl2533: This has to be simplified, I'd start with reverting logic of `if (pageCount > MAX_PAGE_BUTTONS)` we can do `if (pageCount <= MAX_PAGE_BUTTONS) return;` which remove one level of if nesting.

Also, it doesn't work, if I go to page 6 it shows only pages 1-6 and I can't go any further since it doesn't show page 7, same thing if I go from backwards on page 15 it show pages 15-20 and no page 14

src/components/Datatable/Pagination/PageButton.tsx
```diff
@@ -0,0 +1,50 @@
...
+  getFontSize,
+  getFontWeight,
+  getLineHeight,
+  pxToRem,
+} from '../../../utils/helpers';
+import { Button } from '../../Button';
+import { ButtonVariants } from '../../Button/Button.enums';
+import { PageButtonProps } from './Pagination.types';
+
+const PageButton = styled(Button)
```
ajkl2533, line 15: Page buttons are missing border-radius. 
Active button probably shouldn't have any hover effect and shouldn't be rendered as `button` tag. 
When I hover on page button it's missing border from Figma

src/components/Datatable/Pagination/Pagination.tsx
youssefssc: I would prefer to keep the names in this case because they come directly from react-table.

src/components/Datatable/Pagination/NavButton.tsx
youssefssc: I am basically using it so I don't repeat code passing down iconName prop

src/components/Datatable/Pagination/Pagination.tsx
ajkl2533: Pagination is our Component and to be consistent it should follow the same practices as other components so handlers should be prefixed with `on*` we can move the pagination pattern outside of Datatable component in the future and we would need to refactor this

src/components/Datatable/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,206 @@
...
+          pageCount - 1,
+        ]);
+      }
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [pageIndex, pageCount]);
+
+  return (
+    <PaginationWrapper>
+      {isLoading && (
```
ajkl2533, line 115: we should give the loading dedicated space even if it is not shown, because the page numbers jumping around when the spinner is visible.

src/components/Datatable/Pagination/NavButton.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+  background-color: ${({ isDisabled }) =>
+    getColor(isDisabled ? 'transparent' : 'graphite5H')};
+  background-color: ${({ isDisabled }) =>
```
ajkl2533, line 23: Why do we have the `background-color` twice here?

src/components/Datatable/Pagination/PageButton.tsx
```diff
@@ -0,0 +1,51 @@
...
+  .attrs(() => ({
+    variant: ButtonVariants.text,
+  }))
+  .withConfig<PageButtonProps>({
+    shouldForwardProp: (prop) => prop !== 'isActive',
```
ajkl2533, line 20: Instead of `shouldForwardProp` you can use transient props notation (https://styled-components.com/docs/api#transient-props) so change `isActive` to `$isActive` and it won't be forwarded into the underlying component

src/components/Datatable/Pagination/NavButton.tsx
youssefssc: copy/paste mistake, thanks for noticing

src/components/Datatable/Pagination/PageButton.tsx
youssefssc: I kept it flexible because in some cases we can have high page numbers in thousands or tens of thousands in case of a million IPs

src/components/Datatable/Pagination/Pagination.tsx
PrincAm: Our components, our rules 🙂

src/components/Datatable/Pagination/PageButton.tsx
youssefssc: I added a fixed width for spinner so no more horizontal jumping but there are still vertical jumps because of variation in the table cells' heights.

src/components/Datatable/Pagination/PageButton.tsx
youssefssc: active page is now rendered as div and `isActive` prop removed

youssefssc: There is no hover state for nav buttons in Figma so I added the border to highlight it because its color as the same as page buttons on hover

src/components/Datatable/Pagination/Pagination.tsx
youssefssc: @PrincAm  @ajkl2533 There is a total of 4 cases as you can see in the Figma file ... working to simplify.
About the case of page 6, I agree with Radek, I will override the design and show 
[ 1 ...   5  **6**  7  8  ... N] same with 35 or N-5

ajkl2533: It's actually very intresting problem, I'm currently working on my own solution :D

youssefssc: simplified in last commit

src/components/Datatable/Pagination/PageButton.tsx
ajkl2533: https://www.figma.com/file/RWVTqwFZk0NepsI3mcZZFD/Platform-Tables?node-id=1058%3A9957

ajkl2533: Oh you mean Prev/Next buttons, use the same hover effect as `outline` buttons use, it's almost the same styling

src/components/Datatable/Pagination/NavButton.tsx
ajkl2533: You don't need the iconName from IconButton prop at all, just use `Icon` component and add prop `direction` with values `prev` and `next` and decide which icon to show.

src/components/Datatable/Pagination/PageButton.tsx
```diff
@@ -0,0 +1,55 @@
...
+const PageButton = styled(Button).attrs(() => ({
+  variant: ButtonVariants.text,
+}))`
+  &&& {
+    display: inline-block;
+    text-align: center;
+    font-size: ${getFontSize('md')};
+    font-weight: ${getFontWeight('regular')};
+    line-height: ${getLineHeight('lg')};
+    border-radius: ${pxToRem(4)};
```
ajkl2533, line 22:
```suggestion
    border-radius: ${getBorderRadius};
```

src/components/Datatable/Pagination/PageButton.tsx
```diff
@@ -0,0 +1,55 @@
...
+  }
+`;
+
+export const ActivePage = styled.div`
+  display: inline-block;
+  text-align: center;
+  font-size: ${getFontSize('lg')};
+  font-weight: ${getFontWeight('bold')};
+  line-height: ${getLineHeight('lg')};
+  border-radius: ${pxToRem(4)};
```
ajkl2533, line 47:
```suggestion
  border-radius: ${getBorderRadius};
```

src/components/Datatable/Pagination/Pagination.tsx
```diff
@@ -53,45 +51,35 @@ const Pagination: React.FC<PaginationProps> = ({
   canNextPage,
   pageIndex,
   isLoading,
+  numPageButtons = 8,
 }) => {
   const generatePages = (start, end) =>
```
ajkl2533, line 56: This method should go outside of the component, it's pure function no need to recreate on every render

src/components/Datatable/Table/Table.tsx
```diff
@@ -364,6 +364,7 @@ const Table = <D extends Record<string, unknown>>({
         canNextPage={canNextPage}
         canPreviousPage={canPreviousPage}
         isLoading={isLoading}
+        numPageButtons={8}
```
ajkl2533, line 367: You have default value in Pagination component no need to pass it from here

src/components/Datatable/Pagination/Pagination.tsx
```diff
@@ -150,7 +135,6 @@ const Pagination: React.FC<PaginationProps> = ({
         <PageInputLabel> Go to page </PageInputLabel>
         <PageInput
           aria-label="Input"
```
ajkl2533, line 137:
```jsx
  aria-label="Go to Page #"

File: src/components/Datatable/Pagination/Pagination.tsx [hunk 23]
```diff
@@ -0,0 +1,161 @@
...
+const PageInputLabel = styled(Label)`
+  margin: ${pxToRem(6, 16)};
+`;
+const PageInput = styled(StyledNumber)`
+  width: ${pxToRem(72)};
+  height: ${pxToRem(32)};
+  padding: ${pxToRem(8, 16)};
+`;
+
+const Pagination: React.FC<PaginationProps> = ({
```
ajkl2533, line 45: I'm missing the shortcut for condition `(pageCount <= numPageButtons)`. We can bail out from early without need to do the calculations

src/components/Datatable/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,161 @@
...
+  const endPage =
+    currentPage <= threshold
+      ? numPageButtons - 2
+      : currentPage > pageCount - threshold
+      ? pageCount - 1
+      : currentPage + Math.floor((numPageButtons - 4 - 1) / 2);
+
+  const middlePageButtons = generatePages(startPage, endPage);
+
+  const renderPageButton = (page) => {
```
ajkl2533, line 81: this method can be taken out of the component

youssefssc: at the cost of adding two variables to it but no problem ...

ajkl2533 (approved):
LGTM

PR #85, opened 2021-02-17
feat(TagsInput): add component

PrincAm:
Reference https://medium.com/@jerrylowm/build-a-tags-input-react-component-from-scratch-1524f02acb9a


ajkl2533 (approved):
LGTM

src/components/Filters/components/TagsInput/TagsInput.tsx
```diff
@@ -0,0 +1,167 @@
...
+  const [input, setInput] = useState('');
+
+  const handleRemoveTag = (index) => {
+    const newTags = remove(index, 1, tags);
+    onChange(newTags);
+  };
+
+  const handleKeyDown = (event) => {
+    const { value } = event.target;
+    if (event.key === 'Enter' && value) {
```
youssefssc, line 117: I would add here the option to use Tab as well as Enter to create a tag

youssefssc (approved):
LGTM

src/components/Filters/components/TagsInput/TagsInput.tsx
PrincAm: Clarified with UX, it should be only on pressing enter since tab is switching focus to another element

PR #92, opened 2021-02-22
fix(DateRange): update format and condition in renderComponent

PrincAm:


src/components/Filters/FilterRow/FilterRow.tsx
```diff
@@ -89,7 +88,8 @@ const getOperatorOptions = curry((operatorValue) =>
 
 const getFieldOptions = map(normalizeOptions);
 
-const isArrayOfOptionObjects = when(isArray, pipe(head, has('value')));
+const isArrayOfOptionObjects = (value) =>
+  isArray(value) && pipe(head, has('value'))(value);
```
PrincAm, line 92: @ajkl2533 Could be improved? I need to return false if argument is not array

ajkl2533: you can do
```
const isArrayOfOptionObjects = R.both(
 RA.isArray,
 R.pipe(R.head, R.has('value'))
)
```
According to Ramda docs
```
R.both
...
Note that this is short-circuited, meaning that the second function will not be invoked if the first returns a false-y value. ...

File: src/components/Filters/components/DateRangePicker/DateRangePicker.tsx [hunk 2]
```diff
@@ -35,6 +35,7 @@ const DateRangePicker: React.FC<DateRangePickerProps> = ({
       <DatePicker
         calendarClassName="DateRangePicker-calendar"
         className="DateRangePicker-input"
+        dateFormat="d MMM, yyyy"
```
ajkl2533, line 38: So what are the valid input formats? Can I use different formats or just this one?

src/components/Filters/FilterRow/FilterRow.tsx
PrincAm: `R.both` this is it! Perfect

src/components/Filters/components/DateRangePicker/DateRangePicker.tsx
PrincAm: `d MMM, yyyy` is display format and you can input any of these: `MM/DD/YYYY, YYYY/MM/DD, feb 2021, 12 feb 2021, 2021 feb 12, 2021 12 feb`. Parsing itself is done by library so it is much less work than I thought. Even more, we found out Atlas is using same library and ours is most likely first date picker in platform. The refactoring paid off in the end.

ajkl2533 (approved):

PR #95, opened 2021-02-23
fix(Datatable): pass filters when paginating

PrincAm:


src/components/Datatable/Datatable.tsx
```diff
@@ -86,14 +86,29 @@ const Datatable = <D extends Record<string, unknown>>({
...
     any(propEq('isApplied', true), filtersState),
   );
+  const [appliedFilters, setAppliedFilters] = useState(filtersState);
 
   useEffect(() => {
     setPageCount(Math.ceil(totalDataSize / defaultPageSize));
   }, [totalDataSize, defaultPageSize]);
 
+  const handleOnDataFetch = useCallback(
+    (pageIndex, pageSize, sortBy, query) => {
```
ajkl2533, line 96: maybe remove the `query` argument, it won't be passed from the Table either cause the same reasons

src/components/Datatable/Table/Table.tsx
```diff
@@ -324,13 +324,12 @@ const Table = <D extends Record<string, unknown>>({
...
   useEffect(() => {
-    fetchData({
+    fetchData(
       pageIndex,
       pageSize,
       sortBy,
-      filters: [], // TODO: get filters state from context
-      query: '', // TODO: get search query from context
-    });
+      '', // TODO: get search query from context
```
ajkl2533, line 331: remove the query argument

ajkl2533 (approved):
LGTM

PR #98, opened 2021-02-23
fix(layout): pass all props to Col and Row

### LINKED ISSUE #88: styled(Col)`some-styles` does not work
https://github.com/securityscorecard/design-system/blob/26ec617d6f08d81117fa4f0125ac155a144ccfb7/src/components/layout/Col/Col.tsx#L27

I don't know if this is intentional but we can't add styles to a Col component for instance to align content to right, seems annoying to have to add an extra child component to do this. Looks like "className" prop is hidden. I mean something like this could work: 
```
const Col: React.FC<ColProps> = styled(Box).attrs(({ children, cols, offset }) => ({
  ml: `${(100 / 12) * offset}%`,
  ...getColWidth(cols),
  children,
})`
  padding-left: ${getColPadding};
  padding-right: ${getColPadding};
`;
```

PrincAm:
Closes #88

### DIFF
--- a/src/components/layout/Col/Col.tsx
+++ b/src/components/layout/Col/Col.tsx
@@ -24,8 +24,8 @@ const StyledCol = styled(Box)`
   padding-right: ${getColPadding};
 `;
 
-const Col: React.FC<ColProps> = ({ children, cols, offset }) => (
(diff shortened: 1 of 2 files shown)


ajkl2533 (approved):
LGTM

PR #99, opened 2021-02-23
feat(SearchBox): make Suggestions optional

youssefssc:
https://zitenote.atlassian.net/browse/FEP-1625

- on enter - search for given input
- clear button (x) clears the search
- possible to submit empty (clearing previous search)

I connected the `onSearch` functionality to Datatable via `handleOnSearch`

I also put aside the case for suggestions. It should be handled in another PR.


src/components/forms/SearchBar/SearchBar.test.tsx
```diff
@@ -53,7 +54,7 @@ describe('SearchBar', () => {
 
     expect(searchInput.value).toBe('');
   });
-
+  /*
```
PrincAm, line 57: Do we plan to fix them or should be removed?

src/components/forms/SearchBar/SearchBar.tsx
```diff
@@ -80,10 +81,10 @@ const SearchBar: React.FC<SearchBarProps> = ({
     const queryValue = searchInputRef.current.value;
     setSearchPerformed(false);
 
-    if (queryValue.length > 2) {
+    if (hasSuggestions && queryValue.length > 2) {
       setIsSearching(true);
-      const suggestions = await onSearch(queryValue);
-      setSearchResults(suggestions);
+      // const suggestions = await onSearch(queryValue);
```
PrincAm, line 86: Commented code

src/components/Datatable/Datatable.tsx
```diff
@@ -120,23 +124,41 @@ const Datatable = <D extends Record<string, unknown>>({
...
+        query: searchQuery,
+        filters: appliedFilters,
       });
     },
-    [appliedFilters, onDataFetch],
+    [defaultPageSize, onDataFetch, searchQuery, appliedFilters],
   );
 
-  const handleOnFiltersAppply = useCallback(
+  const handleOnFiltersApply = useCallback(
```
ajkl2533, line 146: :D good catch

src/components/forms/SearchBar/SearchBar.tsx
youssefssc: yes both the comments are to be fixed with suggestions

src/components/Filters/components/Input/Input.tsx
```diff
@@ -42,13 +43,23 @@ export const StyledInput = styled.input`
   }
 `;
 
-const Input: React.FC<InputProps> = ({ value = '', onChange }) => (
-  <StyledInput placeholder="String" value={value} onChange={onChange} />
+const Input: React.FC<InputProps> = ({
+  value = '',
+  onChange,
+  onKeyDown = noop,
```
ajkl2533, line 49: why do we need `onKeyDown`

src/components/forms/SearchBar/SearchBar.tsx
PrincAm: Shouldn't be here performed the search? Otherwise it doesn't make much sense

src/components/forms/SearchBar/SearchBar.stories.tsx
```diff
@@ -24,18 +24,24 @@ playground.parameters = {
...
 playground.argTypes = {
+  hasSuggestions: { control: 'boolean' },
   isDisabled: { control: 'boolean' },
   isInvalid: { control: 'boolean' },
 };
 
-export const DefaultSearchBar: Story = () => (
-  <SearchBar aria-label="SearchBar" {...commonArgs} />
+export const Default: Story = () => (
+  <SearchBar aria-label="SearchBar" hasSuggestions {...commonArgs} />
```
ajkl2533, line 33: I would prefer no to add `hasSuggestions` prop. When you say it's `Default` story I would expect that it shows the state with an absolute minimum required configuration.

src/components/Filters/components/Input/Input.tsx
ajkl2533: Found it, in this case, it's maybe better to use destructured rest of props via `{...props}` instead of picking one of many properties

src/components/forms/SearchBar/SearchBar.tsx
```diff
@@ -134,6 +152,7 @@ const SearchBar: React.FC<SearchBarProps> = ({
         type="text"
         value={query}
         onChange={onChangeInput}
+        {...(hasSuggestions || { onKeyDown: handleKeyDown })}
```
PrincAm, line 155: Shouldn't be here `&&`? Please, what do you want to pass? If just simple `onKeyDown={hasSuggestions ? handleKeyDown : noop}` wouldn't be enough

src/components/forms/SearchBar/SearchBar.tsx
youssefssc: yes and to grab `suggestions` as a prop. Right now it's not functional, really just a placeholder

src/components/forms/SearchBar/SearchBar.tsx
youssefssc: the onKeyDown prop is passed only when hasSuggestions is false
`!X && Y` is  the same as `X || Y`

src/components/Filters/components/Input/Input.tsx
youssefssc: I am not sure I understand where the destructing would happen here

src/components/forms/SearchBar/SearchBar.tsx
PrincAm: Understand 🙂

src/components/forms/SearchBar/SearchBar.tsx
youssefssc: adding TO DO comments to both :)

src/components/Filters/components/Input/Input.tsx
ajkl2533: In this `Input` component
```
const Input: React.FC<InputProps> = ({
  value = '',
  onChange,
  ...props
}) => (
  <StyledInput
    placeholder="String"
    value={value}
    onChange={onChange}
    {...props}
  />
);

ajkl2533 (approved):

PrincAm (approved):

PR #100, opened 2021-02-23
feat(Tooltip): use portals for tooltips

ajkl2533:
Closes FEP-1623


src/components/Tooltip/hooks/useTooltip.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  useEffect(() => {
+    setStyle(getPlacementStyles());
+  }, [getPlacementStyles, isShow]);
+
+  useEffect(() => {
+    scrollListener.current = () => {
+      hide();
+    };
+    if (isListenersAdded.current && isNotNull(scrollListener.current)) {
```
PrincAm, line 35: Pls, just check `isNotNull(scrollListener.current)` is not enough? Wondering if `isListenersAdded` ref is needed or what use case I'm missing

PrincAm (approved):
Nice. GJ!

src/components/Tooltip/hooks/useTooltip.tsx
ajkl2533: The problem is we want to remove event listener from window, before we add another one, but if you try to remove event listener which doesn't exist it throws. So this we we check if there is any previous listener which we can remove, if not we just skip the removing part

PrincAm: Check of previous listener, got it

PR #102, opened 2021-02-25
fix(Filters): various tweaks and fixes

PrincAm:


src/components/Filters/Filters.tsx
```diff
@@ -35,28 +36,42 @@ const FiltersContainer = styled(FlexContainer)`
...
-  state = getDefaultState(fields),
+  state,
   onApply,
   onChange,
   onClose,
   onCancel,
   isLoading = false,
 }) => {
-  const [filtersValues, setFiltersValues] = useState<Array<Filter>>(state);
+  const [filtersValues, setFiltersValues] = useState<Array<Filter>>(undefined);
```
ajkl2533, line 46: 'null' would be more appropriate for default state. 'undefined' is more generic and you don't know if it was setup by you or by the app. On the other hand when you use 'null' you are pretty sure it wasn't generated by application

ajkl2533 (approved):

src/components/Filters/Filters.tsx
PrincAm: Good point! It is a  good practice, will change it

src/components/Filters/Filters.tsx
```diff
@@ -35,28 +36,42 @@ const FiltersContainer = styled(FlexContainer)`
...
   const [isDefaultState, setIsDefaultState] = useState(true);
   const [hasUnappliedFilters, setHasUnappliedFilters] = useState(false);
 
   useEffect(() => {
```
youssefssc, line 50: There are 3 useEffects here with shared dependencies. I understand we want to cover all the cases but is it possible to simplify?

PrincAm: I understand you feel something suspicious 🙂 There are 3 useStates and each has own useEffect to set the state. If I would merge all useEffects into one with many if elses then some states would be updated even if not necessarily. I believe it is correct and I kind of like how it is separated, for me it is more clear.

PrincAm: I've been considering if I shouldn't move it into custom hook

youssefssc: My concern is that `fields` is a dependency of the first two useEffects, but I guess fields is a required prop and never changes. If that's correct, then it works :)

PR #104, opened 2021-03-01
fix(Datatable): fix selection using old selected state

ajkl2533:


src/components/Datatable/Datatable.tsx
```diff
@@ -172,20 +176,31 @@ const Datatable = <D extends Record<string, unknown>>({
     [onRowsSelect],
   );
 
+  const memoizedTableConfig = useDeepCompareMemo(
```
PrincAm, line 179: Good optimization! What is the problematic dependency, `restTableConfig`?

ajkl2533: It's one object, so that was the primary issue

youssefssc:
is it my setup or is select/deselect actually slow?

src/components/Datatable/Datatable.tsx
```diff
@@ -172,20 +176,31 @@ const Datatable = <D extends Record<string, unknown>>({
...
+        value={{
+          totalLength: totalDataSize,
+          selectedIds,
+          selectedLength: selectedIds.length,
+          defaultHiddenColumns,
+          defaultColumnOrder,
+          hasExclusionLogic,
+          setHasExclusionLogic,
+          hasSelection,
+          setShouldResetSelectedRows,
```
PrincAm, line 201: Pls, who is the consumer of `setShouldResetSelectedRows`?

PrincAm (approved):
LGTM, hope it works as expected 😉

src/components/Datatable/Datatable.tsx
PrincAm: Yep, I wasn't sure if same handler fn also wasn't causing unnecessary call

src/components/Datatable/Datatable.tsx
ajkl2533: The consumer is `ElementCounter`, this handles situation when you have some selected items and choose `Select none` from the dropdown menu. It wasn't working before, now the `Table` component gets notified through this prop that it should reset selection

ajkl2533:
@youssefssc nope it's slow as hell, I'll work on it later, we need to stabilize functionality now and I'll try to fix performance when we have time

youssefssc (approved):

src/components/Datatable/Datatable.tsx
PrincAm: Right, I missed `ElementCounter`, see it now. Thanks for explanation

PR #106, opened 2021-03-02
feat(Filters): support additional components validations

PrincAm:


src/components/Filters/mocks/options.ts
```diff
@@ -14,7 +14,19 @@ import { Operators } from '../Filters.enums';
...
   {
     conditions: [
-      { component: Input, value: 'is', label: 'is' },
+      {
+        component: {
+          component: Input,
+          props: {
+            maxLength: 15,
+            pattern: '[a-zA-Z]+',
+            patternMessage: 'Use only characters',
```
vaclav-zeman, line 23: This works but I'm just curious if you have considered leaving the validation upon the DT user. Like passing `validate = (values) => { [key]: error }` to the whole <DataTable /> component or individual `validate: requiredFn` as a prop here. 

To me it seems more flexible to pass that responsibility to the user.

PrincAm: I understand, it makes sense. I've leverage HTML 5 input validations like pattern, maxLength for type text and min, max for type number. Then evaluation is not in our hands and also error message is browser's native. If we would change it to validation fn, it would require covering also those. Now, I'm wondering how needed change it is

vaclav-zeman: Yep, html5 attributes are a good approach and don't want it to sound to delete this straight away 👍 

Things to consider though:
- the only problematic field can be IPs or domains since they can be lists afaik? But that's maybe solvable with some crazy pattern
- global validation would allow making more sophisticated validations - like don't allow filter A if filter B is set (probably not something we want atm but who knows)
- adding a new validation type would require to update DS instead of just updating user implementation
- users can reuse validation function from other places 
- UX would want the error messages/styles custom, wouldn't they?

src/components/Filters/Filters.tsx
```diff
@@ -20,23 +28,34 @@ import { Operators } from './Filters.enums';
 const generateId = ({ operator, field, condition }, index) =>
   `${operator}-${field}-${condition}-${index}`;
 
-const getDefaultCondition = (fields) =>
-  (
+const getDefaultConditionAndValue = (fields) => {
```
ajkl2533, line 31:
```js
const getDefaultConditionAndValue = ({conditions}) => { // you don't need whole `fields` array so pass just the first item and take only conditions key from it
  const defaultCondition  = find(propEq('isDefault', true), conditions)
  
  const {
    value: defaultConditionValue, 
    component: defaultConditionComponent
  } = defaultTo(head(conditions), defaultCondition)

  const componentDefaultValue = R.path(['props', 'defaultValue'], component);
  const defaultValue = R.propOr(componentDefaultValue, 'value', componentDefaultValue);

  return { condition: defaultConditionValue, value: defaultValue };
};

File: src/components/Filters/Filters.tsx [hunk 3]
```diff
@@ -20,23 +28,34 @@ import { Operators } from './Filters.enums';
...
+    fields[0].conditions[0];
+
+  const defaultValue =
+    path(['component', 'props', 'defaultValue', 'value'], defaultCondition) ||
+    path(['component', 'props', 'defaultValue'], defaultCondition);
+
+  return { condition: prop('value', defaultCondition), value: defaultValue };
+};
+
+const getDefaultState = (fields) => {
```
ajkl2533, line 43:
```suggestion
const getDefaultState = ([firstField]) => {
  const { condition, value } = getDefaultConditionAndValue(firstField);
  
  return [
    {
      operator: Operators.and,
      field: firstField.value,
      ...
```

src/components/Filters/mocks/options.ts
PrincAm: UX would appreciate to have custom styles of messages and also to have it unified with DS forms. I would merge this version, create follow-up and  make unified form validations across whole DS (platform). We will need also UX input how to style it. How does it sounds to you?

src/components/Filters/Filters.tsx
PrincAm: 👍 Better, shorter composition. Only the initial parameters destructuring helped a lot.

src/components/Filters/mocks/options.ts
vaclav-zeman: Sounds good!

ajkl2533 (approved):

PR #109, opened 2021-03-02
feat(Datatable): connect filter suggestions to SearchBar

youssefssc:
https://zitenote.atlassian.net/browse/FEP-1642

### DIFF
--- a/src/components/Datatable/Datatable.tsx
+++ b/src/components/Datatable/Datatable.tsx
@@ -24,6 +24,7 @@ import {
 } from './Datatable.types';
 import { ControlModule } from './ControlModule';
 import { BatchModule } from './BatchModule';
+import { FilterSuggestion } from '../forms/SearchBar/SearchBar.types';
 
 const StyledDatatable = styled(FlexContainer)`
   border: 1px solid ${getColor('graphiteH')};
@@ -78,6 +79,7 @@ const Datatable = <D extends Record<string, unknown>>({
   dataPrimaryKey,
   onDataFetch = noop,
   isDataLoading = false,
+  onSuggestionsFetch = noop,
   columns,
   tableConfig = {},
   controlsConfig = {},
@@ -104,6 +106,10 @@ const Datatable = <D extends Record<string, unknown>>({
 
   const { state: filtersState = [], onApply: onFiltersApply } = filtersConfig;
   const [searchQuery, setSearchQuery] = useState<string>('');
+  const [filterSuggestions, setFilterSuggestions] = useState<
+    FilterSuggestion[]
+  >([]);
+
   const [pageCount, setPageCount] = useState<number>();
   const [selectedIds, setSelectedIds] = useState<string[]>([]);
   const [hasExclusionLogic, setHasExclusionLogic] = useState<boolean>(false);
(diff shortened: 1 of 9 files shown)


src/components/Datatable/Datatable.tsx
```diff
@@ -134,17 +140,28 @@ const Datatable = <D extends Record<string, unknown>>({
...
   const handleOnSearch = useCallback(
-    (query: string) => {
-      setSearchQuery(query);
-      onDataFetch({
-        pageSize: defaultPageSize,
-        pageIndex: 0,
-        query,
-        filters: appliedFilters,
-      });
+    async (queryValue: string) => {
```
ajkl2533, line 151: why `async` there is no `await` in the function

ajkl2533: why did you change it back to `queryValue`?

src/components/Datatable/Datatable.tsx
```diff
@@ -134,17 +140,28 @@ const Datatable = <D extends Record<string, unknown>>({
...
+      setSearchQuery(queryValue);
+      if (controlsConfig.searchConfig?.hasSuggestions === true) {
```
ajkl2533, line 153: Do not use `controlsConfig` (passed as prop into Datatable component), use only the merged configs from L88 and L96, this would lead to unnecessary bugs like before

src/components/Datatable/Datatable.tsx
```diff
@@ -134,17 +140,28 @@ const Datatable = <D extends Record<string, unknown>>({
...
+        handleOnSuggestionsFetch(queryValue);
+      } else {
+        onDataFetch({
+          pageSize: defaultPageSize,
+          pageIndex: 0,
+          query: searchQuery,
```
ajkl2533, line 159: it can't use the `searchQuery`  it leads to that it uses old state, there needs to be used the arg which is passed into this function, this was already fixed once

src/components/Datatable/Datatable.tsx
```diff
@@ -134,17 +140,28 @@ const Datatable = <D extends Record<string, unknown>>({
...
+          filters: appliedFilters,
+        });
+      }
     },
-    [defaultPageSize, onDataFetch, appliedFilters],
+    [onDataFetch, defaultPageSize, searchQuery, appliedFilters, controlsConfig],
```
ajkl2533, line 164: remove `searchQuery` and do not use objects (`controlsConfig`) in dependency array, rather destructure and use simple types, cause any change of object will cause re-initialization of this function

ajkl2533: also you are missing `handleOnSuggestionsFetch` in dependency array

src/components/Datatable/Datatable.tsx
```diff
@@ -286,6 +304,7 @@ Datatable.propTypes = {
   }),
   batchActions: PropTypes.arrayOf(ActionPropType),
   onDataFetch: PropTypes.func,
+  onSuggestionsFetch: PropTypes.func,
```
ajkl2533, line 307: this property belongs more to `controlsConfig` prop not to the Datatable root props

src/components/Datatable/Datatable.tsx
youssefssc: the await is on L147

src/components/Datatable/Datatable.tsx
youssefssc: fixed

src/components/Datatable/Datatable.tsx
youssefssc: this makes sense. I moved it to controlsConfig on L35 Datatable.types because `searchConfig?: SearchBarProps;`

src/components/Datatable/Datatable.tsx
ajkl2533: but that's is different function, no need to have async here

src/components/Datatable/Datatable.tsx
```diff
@@ -55,6 +56,7 @@ const defaultControlsConfig: ControlsConfig<Record<string, unknown>> = {
     placeholder: 'Search',
     onSearch: noop,
   },
+  onSuggestionsFetch: noop,
```
ajkl2533, line 59: Why not in `searchConfig` it is only useful for the search feature not for the whole controls

src/components/Datatable/Datatable.tsx
ajkl2533: you can extend SearchBarProps, like `SearchBarProp & { onSuggestionsFetch: ... }` or `interface xxx extends SearchBarProps { onSuggestionsFetch: ... }`. it belong to `searchConfig`

ajkl2533 (approved):

PR #110, opened 2021-03-02
fix(Datatable): fix initial load UI

ajkl2533:


src/hooks/useWhyDidYouUpdate.ts
```diff
@@ -0,0 +1,38 @@
+import { useEffect, useRef } from 'react';
+
+export const useWhyDidYouUpdate = (
```
PrincAm, line 3: Cool, for future debugging

PrincAm (approved):
LGTM

PR #114, opened 2021-03-03
fix(Filters): export components

PrincAm:


src/components/Filters/index.ts
```diff
@@ -1 +1,2 @@
+export * as Components from './components';
```
vaclav-zeman, line 1: Hmmmm, wouldn't this be imported as `import { Components } from `@securityscorecard/design-system`? I'd suggest naming it with some Filter prefix

PrincAm: I expect it should be exported like 
```
import { Filters } from '@securityscorecard/design-system'
const { Components } = Filters
```
@ajkl2533 But `import { Components } from '@securityscorecard/design-system/Filters'` should also work? Not to import whole Filters. I could move it one level up to make it easier to use

ajkl2533: Nope it will be exactly as @vaclav-zeman wrote :D 
```suggestion
export * as FilterInputs from './components';
```

ajkl2533: I have no idea how to achieve that `import { Components } from '@securityscorecard/design-system/Filters'` way if you have any suggestion let me know,

PrincAm: I see! Will rename `Components`

ajkl2533 (approved):

PR #115, opened 2021-03-04
feat(Table): style row action button

ajkl2533:
Closes FEP-1624


youssefssc (approved):
checked in detail and LGTM :)

One minor thing I noticed in the action button box in the table header (which should be a square) is that the table header height is 54px not 48px like in design

ajkl2533:
@youssefssc that could be caused by that we still didn't migrate to the new typography scale in implementation but designs are already migrated

PR #117, opened 2021-03-04
fix(SearchBar): disable frequent renders on search change

youssefssc:
https://zitenote.atlassian.net/browse/FEP-1653

- debounce search query 500 ms


src/components/forms/SearchBar/SearchBar.tsx
```diff
@@ -75,33 +76,35 @@ const SearchBar: React.FC<SearchBarProps> = ({
...
   const searchInputRef = useRef<HTMLInputElement>(null);
+  const [query, setQuery] = useState(defaultValue);
   const [isSearching, setIsSearching] = useState(false);
   const [searchPerformed, setSearchPerformed] = useState(false);
   const isMounted = useMounted();
 
-  const performSearch = async (query) => {
+  const debouncedQuery = useDebounce(query, 500);
+
+  const maxLength = 100;
```
ajkl2533, line 86: I wouldn't introduce this validation in fix commit, also it's not a good idea to hardcode the maxLength.

ajkl2533, line 86: Missed that before, but maxLength shouldn't be hardcoded in DS it should be provided as prop also we should have a way how to disable it

ajkl2533 (approved):

src/components/forms/SearchBar/SearchBar.tsx
youssefssc: @ajkl2533  maxLength is not in specs and not sure if it makes sense to include it. I will remove if that's ok with u.

PR #122, opened 2021-03-10
fix(SearchBar): simpler debounce

youssefssc:
fixes FEP-1658

After multiple attempts to make a `useDebouncedCallback` hook, the existing `useControlledInput` one, and typescript fit together without errors, I decided to simplify the input handling and debounce func using only `setTimeout` and `useState`.


ajkl2533 (approved):

youssefssc:
@ajkl2533 A couple of changes in the last two commits
- I kept the non-debounced search because it's needed for onKeyDown instantly
- used useEffect for suggestions display
- added onClear prop
- updated stories to mock new suggestions
- updated tests to check for onSearch and onClear being called

PR #123, opened 2021-03-16
fix: filters Date picker and Select styles

PrincAm:


src/components/Datatable/Table/Table.tsx
```diff
@@ -396,7 +396,7 @@ const Table = <D extends Record<string, unknown>>({
                   {...column.getHeaderProps({
                     ...column.getSortByToggleProps(),
                     ...(isAnyStickyColumn(column.id) && {
-                      style: { flex: '0 0 auto' },
+                      style: { flex: '0 0 auto', zIndex: 0 },
```
PrincAm, line 399: I've tried to increase z-index of select menu but it didn't help. Functionality and look seem to me be still same, just a few pixels on bottom border were marked as different in snapshot diff.

ajkl2533 (approved):

PR #127, opened 2021-03-29
chore: update repo setup

ajkl2533:


.eslintrc.js
```diff
@@ -111,13 +111,21 @@ module.exports = {
     'fp/no-class': 'error',
     'fp/no-valueof-field': 'error',
     'filenames/match-exported': 'error',
-    'jsx-a11y/label-has-for': 'warn',
+    'jsx-a11y/label-has-for': 'off',
```
PrincAm, line 114: Could we remove `jsx-a11y/label-has-for` rule if we have new `jsx-a11y/label-has-associated-control`?

PrincAm (approved):
LGTM

PR #130, opened 2021-03-29
test(mocks): add mocked data for stories and tests

ajkl2533:


youssefssc (approved):
LGTM. To really test, we must wait to connect to the stories in a coming PR I am guessing

PR #131, opened 2021-03-30
feat(Datatable): add dropdown component

ajkl2533:
In Dropdown, there shouldn't be many changes from the current implementation. But I added unit tests.


PrincAm (approved):
It looks fine to me 👍

PR #132, opened 2021-03-30
feat(Datatable): add BatchActions

ajkl2533:
Again nothing magical, but now I actually passing selected row ids and exclusion logic into the `onClick` handlers. Also, test was added.


youssefssc:
from storybook V2 
Screen Shot 2021-03-30 at 11 26 49 AM

ajkl2533:
@youssefssc how did you get into Docs tab? All docs tabs should be disabled for the internal components

youssefssc (approved):
LGTM

ajkl2533:
@youssefssc the redirect is fixed it shouldn't happen

PR #144, opened 2021-03-31
feat(Datatable): add Table component

ajkl2533:


src/components/DatatableV2/Table/Table.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+const Table = <D extends Record<string, unknown>>({
+  columns,
+  data,
+  dataPrimaryKey,
+  dataSize,
+  isDataLoading,
+  rowActions,
+  NoMatchingDataComponent,
+  NoDataComponent,
```
youssefssc, line 63: Both the default components here are taken from `defaultTableConfig` but they are optional props. Perhaps better to move the default values here?

src/components/DatatableV2/Table/Table.reducer.ts
```diff
@@ -0,0 +1,38 @@
...
+}: {
+  collectFetchParams: (
+    pageIndex: number,
+    pageSize: number,
+    sortBy: SortingRule<D>[],
+  ) => void;
+  scrollToTableTop: () => void;
+}): StateReducerFn<D> {
+  return (newState, action) => {
+    // console.log('New State', newState);
```
PrincAm, line 19: commented code

youssefssc (approved):
Nice!

ajkl2533:
@youssefssc nice catch will fix that

PR #145, opened 2021-04-01
feat(Datatable): add Datatable component

ajkl2533:


src/components/index.ts
```diff
@@ -4,6 +4,7 @@ export * from './Button';
 export * from './Card';
 export * from './Collapsible';
 export * from './Datatable';
```
PrincAm, line 6: @ajkl2533 Do you plan to remove old Datatable from codebase?

PrincAm (approved):
LGTM

src/components/index.ts
ajkl2533: sure I do, after migration

PR #146, opened 2021-04-01
refactor(Datatable): convert to function expressions

ajkl2533:


ajkl2533:
@PrincAm I figured out how to use generics with function expressions :D

PR #152, opened 2021-04-08
feat(Datatable): show scrollbar above table

ajkl2533:
Changes in visual regression tests are caused by removing top padding on table.


youssefssc (approved):
LGTM. there are some missing storyshots updates
Screen Shot 2021-04-08 at 2 40 49 PM

PR #154, opened 2021-04-09
feat(Filters): add custom placeholder and units to input components

PrincAm:


src/components/Filters/FilterRow/FilterRow.tsx
```diff
@@ -58,6 +58,15 @@ const SplitField = styled.div<SplitFieldProps>`
   }
 `;
 
+const Units = styled.span`
```
ajkl2533, line 61: why not `Text` component with `md` size?

src/components/Filters/components/DateRangePicker/DateRangePicker.tsx
```diff
@@ -65,7 +66,7 @@ const DateRangePicker: React.FC<DateRangePickerProps> = ({
         endDate={endDate}
         maxDate={maxDate}
         minDate={startDate || minDate}
-        placeholderText="End date"
+        placeholderText={placeholder || 'End date'}
```
ajkl2533, line 69: Shouldn't we be able to pass separate placeholders for first and second field?

src/components/Filters/components/Input/Input.tsx
```diff
@@ -52,6 +52,7 @@ const Input: React.FC<InputProps> = ({
   maxLength,
   pattern,
   patternMessage,
+  placeholder = 'String',
```
youssefssc, line 55: Another default placeholder than 'String' or an empty one would be better in this case

src/components/Filters/components/Number/Number.tsx
```diff
@@ -24,7 +30,7 @@ const Number: React.FC<NumberProps> = ({ value = '', onChange, min, max }) => {
     <StyledNumber
       max={max}
       min={min}
-      placeholder="Number"
+      placeholder={placeholder}
       step="any"
```
ajkl2533, line 34: We probably should be able to set custom `step` value

src/components/Filters/components/Input/Input.tsx
youssefssc: same with other placeholders: 'Date', 'Number', etc

src/components/Filters/components/DateRangePicker/DateRangePicker.tsx
PrincAm: Originally was planned to have placeholder only for `Number` and I added it to all components. I was aware of this use case for start and end date and I din't want to add exception if it is not required. But it is not big deal

src/components/Filters/FilterRow/FilterRow.tsx
PrincAm: Sure, I forget for `Text`

src/components/Filters/components/Input/Input.tsx
PrincAm: Could you elaborate on that? `String`, `Date`, `Number` etc. are default placeholders defined by design

src/components/Filters/components/Number/Number.tsx
PrincAm: We probably should but we also should display step buttons which are hidden now. It is design requirement. The change would be rather small. I will discuss it with UX and eventually get back to it under separate ticket.

src/components/Filters/components/Input/Input.tsx
youssefssc: These were probably put there by Mate as labels not real practical placeholders

PrincAm: Those are real default placeholders. They're helping to end user to understand what data type can input and if there should be something more specific then developer can define custom placeholder which covers this PR :)

src/components/Filters/components/DateRangePicker/DateRangePicker.tsx
PrincAm: Updated

ajkl2533 (approved):

PR #156, opened 2021-04-15
feat(Filters): validate inputs optionally with external func

youssefssc:
- Field level validation with optional custom `validate` functions and error messages
- Form level validations using `onError` prop for inputs and disabling of `Apply` button in case of errors


src/components/Filters/enums.ts
```diff
@@ -0,0 +1,23 @@
+export const patterns = {
```
ajkl2533, line 1: These patterns should be dropped from DS, we shouldn't export anything related to the actual usage of DS components. And the patterns should be implemented directly in the platform.

ajkl2533: At worst should be moved to mocks

src/components/Filters/components/Input/Input.tsx
```diff
@@ -44,25 +49,37 @@ export const StyledInput = styled.input`
...
+  const [isInvalid, setIsInvalid] = useState(false);
 
+  const handleOnValidate = (event) => {
+    const invalid = onValidate(event, validate, patternMessage);
+    setIsInvalid(invalid);
+  };
+
+  const handleOnChange = (event) => {
     onChange(event);
+    if (isInvalid) {
```
ajkl2533, line 80: Not sure it's a good idea to call validate function after the actual change. I'd say the onChange shouldn't happen when the field is invalid IMO

src/components/Filters/components/Input/Input.tsx
```diff
@@ -44,25 +49,37 @@ export const StyledInput = styled.input`
...
+    `}
 `;
 
 const Input: React.FC<InputProps> = ({
   value = '',
   onChange,
   maxLength,
-  pattern,
-  patternMessage,
+  pattern = patterns.string.pattern,
```
ajkl2533, line 65: I would drop all default values related to patterns and let devs decide what they want and need.

src/components/Filters/enums.ts
PrincAm: @youssefssc I'm not aware of use case to have predefined patterns. I would keep it as it was and define each pattern separately for particular story together with pattern message.

src/components/Filters/components/Input/Input.tsx
PrincAm: `onChange` will change Filters state by wrong value and then Filters state is exposed by another public prop `onChange`, it shouldn't probably happen. I would move it into else after validation check. `Input` is controlled component so if `event.target` will be invalid then `onChange` will not be called and user will not be able to continue to input wrong values. I would say it is fine

src/components/Filters/components/Input/Input.tsx
```diff
@@ -44,25 +49,37 @@ export const StyledInput = styled.input`
...
+  patternMessage = patterns.string.patternMessage,
+  validate = validateDefault,
```
PrincAm, line 67: [minor] Prop should be named `onValidate`. `validateDefault` then evaluate the pattern, right? I would name it `patternValidation` or something more descriptive

src/components/Filters/BottomBar/BottomBar.tsx
```diff
@@ -22,6 +22,7 @@ const BottomBar: React.FC<BottomBarProps> = ({
   isLoading = false,
   hasUnappliedFilters,
   isCancelDisabled = false,
+  isApplyDisabled = false,
```
PrincAm, line 25: On wrong input value, it doesn't look `disabled` is changed to `true` 🤔

src/components/Filters/Filters.tsx
```diff
@@ -96,6 +97,14 @@ const Filters: React.FC<FiltersProps> = ({
     }
   }, [filtersValues, fields]);
 
+  const validateFieldValues = () => {
```
PrincAm, line 100: Do we need it? `validationErrors` is always empty array

src/components/Filters/Filters.tsx
```diff
@@ -96,6 +97,14 @@ const Filters: React.FC<FiltersProps> = ({
...
+    const validationErrors = [];
+    // TO DO: check validation for individual fields values
+    setHasInvalidValues(!!validationErrors.length);
+
+    return validationErrors;
```
PrincAm, line 105: Pls, what was the intention with returning of the array of errors? Can be removed?

src/components/Filters/Filters.tsx
```diff
@@ -255,6 +262,7 @@ const Filters: React.FC<FiltersProps> = ({
       })}
       <BottomBar
         hasUnappliedFilters={hasUnappliedFilters}
+        isApplyDisabled={hasInvalidValues}
```
PrincAm, line 265: Be aware that Filters are using form validation with submit button inside `ButtonBar`. Then there is mismatch of all validations together, your custom and native. We want to get rid of the native validation message. Pls see the screenshot 
Screen Shot 2021-04-21 at 11 21 52

src/components/Filters/Filters.types.ts
```diff
@@ -24,6 +24,7 @@ interface ComponentProps {
   maxDate?: Date;
   units?: string;
   placeholder?: string | DateRangePlaceholderProps;
+  validate?: (target: HTMLInputElement) => string | boolean;
```
PrincAm, line 27: Pls, what is the use case for return of `string` by validation function? I would expect input value in, true or false out

src/components/Filters/components/Input/Input.tsx
```diff
@@ -44,25 +49,37 @@ export const StyledInput = styled.input`
...
   placeholder = 'String',
+  isDisabled = false,
```
PrincAm, line 69: Do we need to disable the input field?

src/components/Filters/helpers.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export const onValidate = (
+  event: React.ChangeEvent<HTMLInputElement>,
+  validate: (target: HTMLInputElement) => boolean | string,
+  patternMessage: string,
+): boolean => {
+  const { target } = event;
+  const error = validate(target);
+  target.setCustomValidity(
+    target.validity.patternMismatch ? patternMessage : '',
```
PrincAm, line 13: We don't want to display native validation message. It should be removed

src/components/Filters/components/Input/Input.tsx
```diff
@@ -44,25 +49,37 @@ export const StyledInput = styled.input`
...
-    if (patternMessage) {
-      const { target } = event;
-      target.setCustomValidity(
-        target.validity.patternMismatch ? patternMessage : '',
-      );
-    }
+  const [isInvalid, setIsInvalid] = useState(false);
 
+  const handleOnValidate = (event) => {
+    const invalid = onValidate(event, validate, patternMessage);
```
PrincAm, line 74: It seems to me to be unnecessarily complicated. It will get the value, call validate, it will return good or bad and we will display error message and invalidate the apply button. Or I am wrong?

src/components/Filters/helpers.ts
```diff
@@ -0,0 +1,16 @@
+export const validateDefault = (target: HTMLInputElement): boolean | string => {
+  return !!target.validity.patternMismatch;
+};
+
+export const onValidate = (
```
PrincAm, line 5: Do we need to use this helper function?

PrincAm (commented):
I left a few comment. Also please notice that styles of state button are broken.
Screen Shot 2021-04-21 at 11 44 58

PrincAm:
@youssefssc Please notice, custom validation should be available for `TagsInput` and maybe other components as well. `Input` and `TagsInput` are needed, the rest we can cover later on.

src/components/Filters/components/Input/Input.tsx
youssefssc: When I tried this, the field value could no longer be updated when invalid. It's because onChange is tied to `handleValueChange` in `FilterRow`. Any suggestions @PrincAm ?

youssefssc:
@PrincAm The rest of components' fields are strict in what the user can type or input (... Count, Integer, Number, Select, and dates ) I wasn't sure about Tags but will add the custom validations for it too.

src/components/Filters/components/Input/Input.tsx
PrincAm: Not even by backspace? If not we have to leave it because we control the state as I wrote

PrincAm: There could be also issue with copy pasting the value, we have to bear it in mind

src/components/Filters/helpers.ts
youssefssc: removed this func in commit before last

youssefssc:
@PrincAm The validation for TagsInput are now in.

src/components/Filters/helpers.ts
youssefssc: If removed, the native validation message still displays on "Apply" but with default message "Please match the requested format". IMO it's better to have the same message as the error.

src/components/Filters/FilterRow/FilterRow.tsx
```diff
@@ -154,9 +163,8 @@ const renderSelectComponent = (Component, value, onChange) => {
   );
 };
 
-const renderComponent = (Component, value, onChange) => {
+const renderComponent = (Component, value, onChange, onError) => {
```
PrincAm, line 166: I think we don't need to pass `onError` to component without additional properties because all validations should be defined in component props object. Or I am missing some use case?

src/components/Filters/FilterRow/FilterRow.tsx
```diff
@@ -166,10 +174,9 @@ const renderComponent = (Component, value, onChange) => {
   }
   // Component with props
   if (typeof Component === 'object' && has('props', Component)) {
-    return renderComponentWithProps(Component, value, onChange);
+    return renderComponentWithProps(Component, value, onChange, onError);
   }
-
-  return <Component value={value} onChange={onChange} />;
+  return <Component value={value} onChange={onChange} onError={onError} />;
```
PrincAm, line 179: I think we don't need to pass onError to component without additional properties because all validations should be defined in component props object. Or I am missing some use case?

src/components/Filters/Filters.tsx
```diff
@@ -58,7 +58,7 @@ const getDefaultState = ([firstField]: Field[]) => {
   ];
 };
 
-const Form = styled.form`
+const Form = styled.div`
```
PrincAm, line 61: Could we rename `Form` to something else if it is not from anymore? Also you can `FlexContainer` instead

src/components/Filters/components/Count/Count.tsx
```diff
@@ -9,23 +11,34 @@ const Count: React.FC<NumberProps> = ({
...
+  const handleOnChange = (event) => {
+    onChange(event);
+    const error = !validateNumber(event.target.value, min, max);
+    setIsInvalid(error);
+    onError(error);
+  };
 
   return (
-    <StyledNumber
-      max={max}
```
PrincAm, line 21: We can't use `min`, `max`, `step` native attributes?

src/components/Filters/components/Count/Count.tsx
```diff
@@ -9,23 +11,34 @@ const Count: React.FC<NumberProps> = ({
...
+  const [isInvalid, setIsInvalid] = useState(false);
+
   const handleKeyPress = (event) => {
     if (!/[0-9]/.test(event.key)) {
       event.preventDefault();
     }
   };
+  const handleOnChange = (event) => {
+    onChange(event);
+    const error = !validateNumber(event.target.value, min, max);
```
PrincAm, line 26: If `error` has boolean type, use `is` or `has` prefix in name

src/components/Filters/components/Count/Count.tsx
youssefssc: They would be useless to add here. The range validation is done in `validateNumber` so we can show our custom errors.

src/components/Filters/components/Input/Input.tsx
```diff
@@ -52,17 +63,23 @@ const Input: React.FC<InputProps> = ({
...
-      const { target } = event;
-      target.setCustomValidity(
-        target.validity.patternMismatch ? patternMessage : '',
-      );
-    }
+  const [isInvalid, setIsInvalid] = useState(false);
+
+  const handleOnValidate = (event) => {
+    const { target } = event;
+    const error = onValidate(target);
```
PrincAm, line 75: Same here, `onValidate` should return boolean and name should have boolean prefix

youssefssc: I updated the names for all inputs but in the Input and TagsInput case we need the whole target because 
```
validatePattern = (target: HTMLInputElement): boolean => {
  return !!target.validity.patternMismatch;
};```

File: src/components/Filters/components/Count/Count.tsx [hunk 20]
```diff
@@ -9,23 +11,34 @@ const Count: React.FC<NumberProps> = ({
   min,
   max,
   placeholder = 'Count',
+  patternMessage = 'Use only integers',
```
PrincAm, line 14: I would rename all `patternMesage`s to something more general like `errorMessage` since it is not bound only to pattern. Maybe will be needed update of FE if we will change the API of filters

src/components/Filters/helpers.ts
```diff
@@ -0,0 +1,16 @@
+export const validatePattern = (target: HTMLInputElement): boolean => {
+  return !!target.validity.patternMismatch;
```
PrincAm, line 2: [minor] You can omit `return`

src/components/Filters/helpers.ts
```diff
@@ -0,0 +1,16 @@
...
+};
+
+export const validateNumber = (
```
PrincAm, line 5: It should be covered by unit test

src/components/Filters/components/Count/Count.tsx
PrincAm: I understand how it works. I thought we agreed we will use native validation and just display custom message for sake of simplicity, so it is not possible?

src/components/Filters/components/Input/Input.tsx
PrincAm: Okey. `onValidate` is not custom validation here, it mislead me a bit.

src/components/Filters/components/Count/Count.tsx
youssefssc: It is possible but does nothing as I understand

PrincAm:
@youssefssc Please notice I can apply filters even if there are some fields invalid
Screen Shot 2021-04-26 at 14 24 33

PrincAm:
@youssefssc One more observation. I would mark empty value as valid to make it consistent because it is same like on the start. Furthermore, if you click on Apply, all filters with empty values are removed.
Screen Shot 2021-04-26 at 14 31 28

youssefssc:
@PrincAm I added fixes for the last two bugs in screenshots. So now empty values are valid and we have an array `validsValues` which must all be `true` to enable the Apply button.  It's not complete yet as I am not sure how and where exactly to update the size of this array when the number of filters change but please take a look and let me know if this is going in the right direction. Thanks.

PrincAm:
@youssefssc We have two places where we keep information about inputs validity. One in each component `isInvalid` and second in Filters `validsValues` (`currentFieldIndex`). We should avoid that and have only one source of truth. I would remove local states from components, kept them simple as they were and move state with errors to `Filters` component similarly how filters values are managed. Filters errors can be object where key will be an index of the row and value boolean, e.g. `{ 0: true, 2: true }`

PrincAm:
@youssefssc Also pls notice there are some prop types warnings when used validation
Screen Shot 2021-04-28 at 12 58 59

src/components/Filters/Filters.tsx
```diff
@@ -97,6 +93,16 @@ const Filters: React.FC<FiltersProps> = ({
     }
   }, [filtersValues, fields]);
 
+  const handleError = (hasError, index) => {
+    // TO DO: handle all validations in FilterRow in handleValueChange
```
PrincAm, line 97: I think this TODO can be removed

src/components/Filters/Filters.tsx
```diff
@@ -97,6 +93,16 @@ const Filters: React.FC<FiltersProps> = ({
...
+    const newValidsValues = validsValues;
```
PrincAm, line 98: It should be a copy

src/components/Filters/Filters.tsx
```diff
@@ -78,6 +72,7 @@ const Filters: React.FC<FiltersProps> = ({
   const [filtersValues, setFiltersValues] = useState<Array<Filter>>(null);
   const [isDefaultState, setIsDefaultState] = useState(true);
   const [hasUnappliedFilters, setHasUnappliedFilters] = useState(false);
+  const [validsValues, setValidsValues] = useState<boolean[]>([true]);
```
PrincAm, line 75: Isn't `validValues` correct name?

PrincAm: Shouldn't be the default value an empty array? Or it should be derived from `state`

PrincAm:
If I remove row with invalid value then following row is marked as invalid. It's not correct
Screen Shot 2021-04-29 at 15 20 49

src/components/Filters/components/TagsInput/TagsInput.tsx
```diff
@@ -127,56 +131,63 @@ const TagsInput: React.FC<TagsInputProps> = ({
...
 
   const handleOnBlur = (event) => {
-    const { target } = event;
-    const { value, validity } = target;
+    const {
+      target: { value },
+    } = event;
 
-    if (value && !validity.patternMismatch) {
+    if (value && !isInvalid) {
```
PrincAm, line 156: I'm able to add invalid tag value on blur :(

PrincAm: But Apply button is disabled so we can fix that later, imo

src/components/Filters/Filters.tsx
youssefssc: the default is an array of one true element because it's one Field / FilterRow. We can move it to the first useEffect but it's the same

src/components/Filters/components/TagsInput/TagsInput.tsx
PrincAm: Story "Tags Input Validation", insert wrong value and click outside the input. Tag with incorrect value is added. You can compare this behavior with adding of wrong value on enter.

youssefssc:
I addressed the new comments and implemented the update for validValues on filter add/remove/clear all
Also removed isDisabled for Input 

@PrincAm Let me know if you think this PR is ready to be merged, and if there is a small bug let me know and I will create a ticket for it

src/components/Filters/Filters.tsx
PrincAm: If `state` passed down to Filters will include 5 "filters" definitions then it should include five `true`s. Therefore it should be derived from state.length

src/components/Filters/components/TagsInput/TagsInput.tsx
youssefssc: I don't have this behavior on blur  :S

PrincAm: It is fixed

PrincAm:
I can't insert correct value when used `min`, `max` validation
Screen Shot 2021-04-29 at 16 56 05

src/components/Filters/components/Input/Input.tsx
```diff
@@ -32,9 +35,9 @@ export const StyledInput = styled.input`
     border-color: ${getFormStyle('focusBorderColor')};
   }
 
-  &:invalid {
-    ${stateStyles}
-    border-color: ${getFormStyle('invalidBorderColor')};
+  &:disabled {
```
PrincAm, line 38: Do we need this styles for disabled state? I don't think we are disabling the input

PrincAm (approved):
LGTM

PR #163, opened 2021-04-21
feat(Datatable): add per page selection option

ajkl2533:
Changes of the visual test are caused by changing the default behaviour of the selection, from now on we have exclusive selection disabled.


youssefssc:
The Batch Module with-selected-rows-and-exclusion-logic storybookshot shows "No Data". It looks like an error here.

ajkl2533:
@youssefssc good catch, pushing update

PR #165, opened 2021-04-27
fix: remove reference of filters state to DT store

PrincAm:
~~I had to make a deep copy of filteringStoreState because it is read only with reference to store. Value is then saved to Filters state and mutated.~~

Filters state is not mutated.


ajkl2533 (approved):
LGTM

PR #168, opened 2021-05-03
feat(ColumnsControls): add dropdown pane for control columns order

ajkl2533:


src/components/Datatable/ControlsModule/ColumnsControls/ColumnsControls.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  const handleOpenColumnsControl = () => {
+    onOpen();
+  };
+  const handleCloseColumnsControl = () => {
+    onClose();
+    reinitializeOrderedColumns();
+  };
+  const handleApplyColumnsControl = () => {
+    onApply(!isInDefaultOrder);
```
vaclav-zeman, line 42: Why not do `if (!isInDefaultOrder) onApply()` instead?

src/components/Datatable/hooks/useColumnsControls.ts
```diff
@@ -0,0 +1,44 @@
...
+      s.columnOrder = defaultColumnOrder;
+    });
+  }, [defaultColumnOrder]);
+
+  useEffect(() => {
+    const unsubscribe = DatatableStore.subscribe(
+      (s) => s.columnOrder,
+      (columnOrder) => {
+        onColumnOrderChange(columnOrder);
+      },
```
vaclav-zeman, line 37:
```suggestion
      const unsubscribe = DatatableStore.subscribe(prop('columnOrder'), onColumnOrderChange);
```

vaclav-zeman (approved):
lgtm, just a few nitpicks

src/components/Datatable/hooks/useColumnsControls.ts
ajkl2533: as much as I agree, but I would need to change it also on all other places to remain consistent :) maybe some next PR

src/components/Datatable/ControlsModule/ColumnsControls/ColumnsControls.tsx
ajkl2533: I need that info to use it in ControlsModule, so this is correct but the intention will be revealed in next PR :)

youssefssc:
I am not sure what are the acceptance criteria of this PR, but it's looks good.
Small note: Without any reordering columns on the story in storybook, using "Reset to defaults" reorders the columns. This doesn't look intentional.

src/components/Datatable/ControlsModule/ColumnsControls/ColumnsControls.tsx
```diff
@@ -0,0 +1,66 @@
...
+  const handleResetColumnsControl = () => {
+    onReset();
+    resetOrderedColumns();
+  };
+
+  return (
+    <ControlDropdown
+      defaultIsControlPaneOpen={defaultIsColumnsControlsOpen}
+      renderHandler={children}
+      title="Columns"
```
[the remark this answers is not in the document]
ajkl2533: Again, if we would ever need to translate we would need to do changes in lots of places. For now, it would add just add unnecessary complexity.

src/components/Datatable/ControlsModule/ColumnsControls/hooks/useColumnOrder.ts
```diff
@@ -0,0 +1,77 @@
+import { useEffect, useState } from 'react';
+import { comparator, keys, lt, sort } from 'ramda';
+import deepEqual from 'fast-deep-equal';
+
+import { DatatableStore } from '../../../Datatable.store';
+
+const compare = (order) => (a, b) => {
```
youssefssc, line 7: an explanation of what the logic is meant to achieve here and a unit test would be helpful.

ajkl2533:
@youssefssc if you mean in the storybook, it's because I'm have set up preferred order in the story

src/components/Datatable/ControlsModule/ColumnsControls/ColumnsControls.stories.tsx
```diff
@@ -0,0 +1,43 @@
...
+  parameters: {
+    viewMode: 'story',
+    docs: { disable: true },
+    previewTabs: { 'storybook/docs/panel': { hidden: true } },
+  },
+} as Meta;
+
+export const Default: Story<ColumnsControlsProps> = (args) => {
+  useColumnsControls(action('onColumnOrderChange'), simpleColumns, [
+    'country',
```
youssefssc, line 22: now I see the custom order @ajkl2533 . So it's set up so that we can pass partial custom order and the rest is automatically filled. What's the reason for this preference?

ajkl2533: correct, this is taken from the react-table which supports this feature. Also that is the logic in the compare function. Default order is taken from the columns definition array.

youssefssc (approved):
LGTM

PR #182, opened 2021-05-21
fix(portals): fix portals stacking

ajkl2533:
Ref FEP-1740

### DIFF
--- a/src/components/FullscreenModal/hooks/useModal.tsx
+++ b/src/components/FullscreenModal/hooks/useModal.tsx
@@ -1,0 +1,43 @@
+import React, { useContext, useEffect, useRef } from 'react';
+import { createPortal } from 'react-dom';
+import { isNull } from 'ramda-adjunct';
+
+import { DSContext } from '../../../theme/DSProvider/DSProvider';
+
+const createEl = (id: string): HTMLElement => {
+  const el = document.createElement('div');
+  el.setAttribute('id', id);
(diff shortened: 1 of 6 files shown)


src/components/FullscreenModal/hooks/useModal.tsx
```diff
@@ -0,0 +1,43 @@
...
+      current.style.position = 'static';
+      current.style.zIndex = 'auto';
+      current.style.top = 'auto';
+      current.style.left = 'auto';
+    };
+  }, [portalsContainerId]);
+
+  return ({ children }) => {
+    if (isNull(containerRef.current)) return null;
+    return createPortal(children, containerRef.current);
```
andrasbudayssc, line 41: since we're only passing it to the portal I think the terminology could be changed to singular ```child``` as it's in the documentation.

andrasbudayssc (approved):
Good stuff! 💯

PrincAm (approved):
Nice!

src/components/FullscreenModal/hooks/useModal.tsx
ajkl2533: Well, it actually returns functional component and `children` is `children` property not sure if it make sense alias it just for fun :D

andrasbudayssc: yep, that was exactly the argument I was having with myself as well 😂

PR #185, opened 2021-05-25
fix(Filters): remove preventing default in components

PrincAm:
We missed this when refactoring Filters from `form` to `div` container. There is no need to keep it there, moreover it was preventing selection in dropdown on enter.


ajkl2533 (approved):
LGTM

PR #187, opened 2021-05-27
feat(Table): reset pagination after sort and filter

andrasbudayssc:
# Reset pagination when filters or sorting has changed

## changes
 - due to having to support both client/server side pagination we have two distinct states for pageIndex and because of this, we had to have the ability to change the page without side effects (`gotoPage` from `react-use-table`) in case the the page number has been changed elsewhere (e.g. filters which is outside the scope of `Table`). 
 - when `sortBy` changes reset page index to 0
 - when applying filters reset page index to 0


ajkl2533 (approved):
LGTM

PR #193, opened 2021-06-01
feat(Datatable): update single select feature

ajkl2533:


src/components/Datatable/Table/Table.styles.ts
```diff
@@ -178,6 +183,8 @@ export const StyledTable = styled.table`
     }
 
     .ds-table-checkbox-label {
```
PrincAm, line 185: Do we support label for select checkbox? :) I can't find it in Storybook. If yes it would worth to add the story

PrincAm: Can we also remove radio styles? They are still used or you want to keep them?

PrincAm (approved):
LGTM, works well! Comment to consider :)

src/components/Datatable/Table/Table.styles.ts
ajkl2533: No, we don't support the label, but label tag is what actually provides the clicking function on custom-styled checkbox:) I'll remove the radio styles, I forgot them

PrincAm: Got it, thanks for explanation

PR #196, opened 2021-06-03
feat(theme): add spacing scale and helpers

ajkl2533:


src/theme/space.stories.mdx
```diff
@@ -0,0 +1,41 @@
...
+
+<Figma url="https://www.figma.com/file/vnzYfrwIISRYkr2iYE5TcD/Platform-Atoms?node-id=2314%3A51" />
+<Figma url="https://www.figma.com/file/vnzYfrwIISRYkr2iYE5TcD/Platform-Atoms?node-id=2314%3A53" />
+
+## Inseting
+
+We have three different kinds of inset:
+  1. `square` - vertical and horizontal inset are equal
+  2. `squish` - vertical inset is half of horizontal
+  3. `stretch` - horizontal inset is half of vertical
```
vaclav-zeman, line 34: Not sure I like this opinionated approach instead of being able to just set `paddingX="xs" paddingY="xxs"`.

But I guess being strict works for us, especially since we have several independent teams

vaclav-zeman (approved):
LGTM 👍

src/theme/space.stories.mdx
ajkl2533: It should be ok as these three concepts will cover the majority of use-cases. If you have some legit use-case for some special combination of padding you are always able to do it in CSS and use whatever value from scale with `getSpace`.
Recommend https://medium.com/eightshapes-llc/space-in-design-systems-188bcbae0d62 where we took inspiration about the spacing scale and those three options.

Once the layout primitives, we should be able to cover most of the components with composition.

PR #197, opened 2021-06-03
Datatable request cancelation feature

andrasbudayssc:
Quite a hefty one, so sorry about that in advance.

## Breaking change
`onCancel` is no longer a part of the `filterConfig` object but is a top level prop on the `Datatable` component named `onCancelLoading`, right next to `onDataFetch`.

## Summary
We need to enable our users to cancel the current, potentially time/resource-consuming request and escape from the possible blocking loading state without having to navigate back and forth or refresh the entire page.

This PR covers the required changes in the design system for the `Datatable` and `Filters` components.

Basically it introduces a new property on the `Table` state which will keep the value of the previous state, and if it detects that the `Datatable` store has been canceled it will reset to that previous state. This allows `Filters` to be able to make meaningful changes in the `Table` scope as a sibling.

## Potential issues
Unfortunately, as the Filter states are duplicated (once in `Datatable` store and once in the local state) and because the `Filters` component is decoupled from the `Datatable` it cannot know when the request was canceled from the `LoadingOverlay` thus the filter will display the applied tick in this case, might be worth disabling the overlay cancel button if filters are opened.


src/components/Datatable/Table/Table.tsx
```diff
@@ -199,17 +231,28 @@ TableProps<D>): React.ReactElement => {
...
-    };
-  }, [setColumnOrder, gotoPage]);
+  useEffect(
+    () =>
+      DatatableStore.subscribe(
+        pick(['filters', 'isCanceled']),
+        when(
+          allPass([
+            propEq('isCanceled', false),
+            compose(
```
ajkl2533, line 241: use `pipe` instead of `compose`

src/components/Datatable/Table/Table.tsx
```diff
@@ -199,17 +231,28 @@ TableProps<D>): React.ReactElement => {
...
-        if (any(propEq('isLoading', true))(filters)) gotoPage(0);
-      },
-    );
-    return () => {
-      unsubscribe();
-    };
-  }, [setColumnOrder, gotoPage]);
+  useEffect(
+    () =>
+      DatatableStore.subscribe(
```
ajkl2533, line 236: Any reason why you not unsubscribing?

src/components/Datatable/hooks/useDataFetch.ts
```diff
@@ -1,21 +1,21 @@
...
-        pageIndex: s.pageIndex,
-        pageSize: s.pageSize,
-        sortBy: s.sortBy,
-        filters: s.filters,
-        query: s.query,
-      }),
-      (params) => {
-        onDataFetch(params);
-      },
+      pick([
```
ajkl2533, line 10: I already mentioned it to someone, I'm definitely in favour to use ramda here, but we should do it consistently for all occurrences. We should be consistent throughout codebase and not to have half done by ramda and the other half differently.

src/components/Datatable/Table/Table.tsx
andrasbudayssc: ah I was afraid this might be called out :) so the thing is, it actually does if you look carefully (it returns the `unsubscribe` function returned from the `subscribe` method but it means it's probably not so easy to read, just wanted to save a couple of lines 😄

ajkl2533: oh I see now, there aren't parenthesis. Yep it's easy to miss

ajkl2533: Again that's fine if we use it consistently throughout codebase, so either stick to current style or change it everywhere

src/components/Datatable/Datatable.stories.tsx
```diff
@@ -150,6 +151,12 @@ export const Playground: Story<DatatableProps<Data>> = (args) => {
     tableConfig: argsTableConfig,
     ...restArgs
   } = args;
+
+  const onCancelLoading = () => {
```
OldrichKruchna, line 155: I'm just wondering why is call `onCancelLoading` and not just `onCancel`? Is there some special cancel loading?

src/components/Datatable/Table/Loading/LoadingOverlay.tsx
```diff
@@ -15,14 +19,22 @@ const overlayStyles = css`
...
     left: 50%;
     top: 50%;
     transform: translate(-50%, -50%);
     margin: 4rem;
     background: ${getColor('graphite5H')};
+    padding: 0 0.5rem;
+    height: 2.5rem;
+    border-radius: 4px;
+
+    > span {
```
OldrichKruchna, line 34: What about to create just a simple `LoadingText` component? That would make this style rule not depending on the HTML structure.

OldrichKruchna (approved):
I guess I don't have many insights to the design system structure, so some things are hard to code review for me, but overall looks as pretty solid solution :)

src/components/Datatable/Table/Table.tsx
andrasbudayssc: yea will revert :)

src/components/Datatable/hooks/useDataFetch.ts
andrasbudayssc: I agree 100%, I'm for Ramda but consistency > personal preferences. I only went with the change here because the entire file is now Ramda based, way I see it, is that if we can get a file completely migrated it's already a win as it's very unlikely we'll ever have the chance to do it all at once. what do you think?

ajkl2533: Can you promise me that you'll migrate all to Ramda? :D I get it could you please create a ticket in the DS epic so we can tackle this and don't forget about it?

src/components/Filters/BottomBar/BottomBar.tsx
```diff
@@ -56,12 +64,23 @@ const BottomBar: React.FC<BottomBarProps> = ({
...
         <Button
           color="primary"
-          disabled={isApplyDisabled}
-          /* disabled={isLoading} TODO enable by https://zitenote.atlassian.net/browse/FEP-1645 */
+          disabled={isApplyDisabled || canCancel}
           variant="solid"
           onClick={onSubmit}
         >
-          Apply
+          {canCancel ? (
```
ajkl2533, line 71: BTW we have loading button state https://securityscorecard.github.io/design-system/alpha/?path=/docs/components-button--playground#loading-buttons

ajkl2533 (approved):
This looks good to me, nice job 👍

src/components/Filters/BottomBar/BottomBar.tsx
PrincAm: Ticket for the update https://zitenote.atlassian.net/browse/FEP-1645 Perhaps is time to tackle it

andrasbudayssc: Good point! I've discovered that in the beginning and then forgot about it due to the discrepancy, so thanks for pointing it out :) Is the current loading state used anywhere besides here? if not, I would create a ticket that covers the migration to the new designs for the general Button component.

andrasbudayssc: never mind @PrincAm beat me to it 🤣 yeah I'll take care of that!

src/components/Filters/Filters.tsx
```diff
@@ -97,20 +102,27 @@ const Filters: React.FC<FiltersProps> = ({
   }, [filtersValues, fields]);
 
   useEffect(() => {
```
PrincAm, line 104: Filters `useEffect`s pretty grew up by last updates including error handling and request cancelation. With adding another logic I would move them to testable custom hook.

PrincAm (approved):
Good job Andras!

src/components/Datatable/hooks/useDataFetch.ts
andrasbudayssc: haha I'll put it up as one of the goals, and absolutely https://zitenote.atlassian.net/browse/FEP-1818

PR #207, opened 2021-06-09
fix(fep-1808): tags are not removed after click on clearall

eremineu:
Problem: Tags are not removed from the select field if the multi-select field is in the first position and "condition" is the same as the default. (example: "detection method" first in the list). 

Solved by adding a "key" to select elements.

I'm not sure if it's the best solution. I'm open to any better suggestions.


ajkl2533 (approved):
LGTM, it just forces to re-render the component. 👍

PR #213, opened 2021-06-15
fix(Datatable): fix in-browser pagination

ajkl2533:
Currently, when the `hasServerSidePagination` set to true
we are still rendering all table rows instead of just the current
page. This PR fixes this behaviour.

We also forgot to reset pagination after search is performed.


src/components/Datatable/Table/Table.tsx
```diff
@@ -232,6 +237,18 @@ TableProps<D>): React.ReactElement => {
     };
   }, [setColumnOrder]);
 
+  useEffect(() => {
+    const unsubscribe = DatatableStore.subscribe(prop('pageIndex'), (index) => {
```
andrasbudayssc, line 241: might be worth considering Ramda

andrasbudayssc (approved):
LGTM

src/components/Datatable/Table/Table.tsx
ajkl2533: done

PR #217, opened 2021-06-21
feat(Nav): deprecate Nav and NavItem components

ajkl2533:
We have decided to deprecate the Nav component in favour of the text variant of the Tabs component. This PR adds info to Storybook and also throws deprecation warning in non-production environments in the console.


jurajhrib (approved):
LGTM

PR #224, opened 2021-06-30
feat(Tabs): add custom pattern matching for selected tab

ajkl2533:
This PR adds an option to pass custom pattern matching predicate to determine which tab is currently selected. 
By default, the currently selected tab is determined by strict equality between `selectedValue` and Tab `value`. 
But we can provide for example `R.startsWith` as a predicate and in that case, when we have `selectedValue = "overview/section"` then Tab with `value= "overview"` will be marked as selected.


abykovssc (approved):

jurajhrib (approved):
LGTM!

PR #226, opened 2021-07-07
feat(Stack): change justify default prop to 'stretch'

ajkl2533:
After using this component in several places it seems that `stretch` is a better default value for `justify` than `flex-start`. Let's change that before we start using LP in prod.


eremineu (approved):
LGTM

PR #227, opened 2021-07-07
Custom width for toast notification

eremineu:


src/components/Toast/Toast.tsx
```diff
@@ -74,10 +74,11 @@ const StyledIcon = styled(Icon)`
 
 const stopPropagation = (event) => event?.stopPropagation();
 
-const Toast: React.FC<ToastProps> = ({ onClose, children }) => (
+const Toast: React.FC<ToastProps> = ({ onClose, children, width }) => (
   <ToastContainer
     alignItems="flex-start"
     justifyContent="space-between"
+    width={width}
```
ajkl2533, line 81: For internal props passed into styled-components please use `$...` prefix. We want to prevent propagation of props into DOM so we are using SC's [Transient Props](https://styled-components.com/docs/api#transient-props)
```suggestion
    $width={width}
```

src/components/Toast/Toast.tsx
```diff
@@ -91,6 +92,11 @@ const Toast: React.FC<ToastProps> = ({ onClose, children }) => (
 
 Toast.propTypes = {
   onClose: PropTypes.func.isRequired,
+  width: PropTypes.number,
+};
+
+Toast.defaultProps = {
```
ajkl2533, line 98: Please drop the `.defaultProps` and set defaults directly in the component, we are using default props only when we export directly styled component.
```js
const Toast: React.FC<ToastProps> = ({ onClose, children, width = 370 }) =>

File: src/components/Toast/Toast.tsx [hunk 3]
```diff
@@ -22,12 +22,12 @@ const ToastFromTop = keyframes`
   }
 `;
 
-const ToastContainer = styled(FlexContainer)`
+const ToastContainer = styled(FlexContainer)<{ width?: number }>`
```
ajkl2533, line 25:
```suggestion
const ToastContainer = styled(FlexContainer)<{ $width?: ToastProps['width'] }>`
```

ajkl2533 (approved):
Couple of small nitpicks but LGTM.

Before merge please squash both commits into one with message:
`feat(Toast): add 'width' property`

PR #231, opened 2021-07-16
fix(TextArea): pass props to correct element

ajkl2533:
Closes FEP-1889


src/components/forms/TextArea/TextArea.tsx
```diff
@@ -135,6 +143,8 @@ TextArea.propTypes = {
   isDisabled: PropTypes.bool,
   value: PropTypes.string,
   defaultValue: PropTypes.string,
+  style: PropTypes.shape({}),
```
youssefssc, line 146: what does styles used for here?

ajkl2533: it's just inline CSS style prop if we need to use it

youssefssc (approved):
LGTM

PR #239, opened 2021-07-29
feat(Datatable): add support for default page index

andrasbudayssc:


youssefssc (approved):
LGTM

src/components/Datatable/Table/Table.tsx
```diff
@@ -139,6 +140,7 @@ TableProps<D>): React.ReactElement => {
       initialState: {
         // PAGINATION
         pageSize: defaultPageSize,
+        pageIndex: defaultPageIndex || 0,
```
ajkl2533, line 143: Maybe would be better to set the default value in https://github.com/securityscorecard/design-system/blob/alpha/src/components/Datatable/defaultConfigs.ts as every other default

PR #242, opened 2021-07-30
feat: extract and export ControlDropdown and Sortable list

ajkl2533:
Closes FEP-1919


youssefssc (approved):
LGTM

src/components/ControlDropdown/ControlDropdown.tsx
```diff
@@ -0,0 +1,129 @@
...
+      }
+    },
+    onHide: onClose,
+    placement,
+    isElevated: true,
+    hasInternalShowHide: false,
+  });
+
+  return (
+    isOpen && (
```
andrasbudayssc, line 61: just to save indentation and readability might be worth checking this outside the JSX context, since it's a top level conditional

src/components/Datatable/Datatable.tsx
```diff
@@ -162,6 +162,8 @@ Datatable.propTypes = {
     }),
     defaultIsFilteringOpen: PropTypes.bool,
     defaultIsFilteringApplied: PropTypes.bool,
+    defaultIsColumnOrderingOpen: PropTypes.bool,
+    defaultIsColumnOrderingApplied: PropTypes.bool,
```
andrasbudayssc, line 166: 🙏 thank you, didn't think of it but will actually need this for the persistent state

src/components/SortableList/SortableItem.tsx
```diff
@@ -0,0 +1,80 @@
...
+          >
+            <path
+              clipRule="evenodd"
+              d="M 9 7 C 9 6.448 9.448 6 10 6 L 14 6 C 14.552 6 15 6.448 15 7 C 15 7.552 14.552 8 14 8 L 10 8 C 9.448 8 9 7.552 9 7 Z M 9 11 C 9 10.448 9.448 10 10 10 L 14 10 C 14.552 10 15 10.448 15 11 C 15 11.552 14.552 12 14 12 L 10 12 C 9.448 12 9 11.552 9 11 Z M 10 14 C 9.448 14 9 14.448 9 15 C 9 15.552 9.448 16 10 16 L 14 16 C 14.552 16 15 15.552 15 15 C 15 14.448 14.552 14 14 14 L 10 14 Z"
+              fill="currentColor"
+              fillRule="evenodd"
+            />
+          </svg>
+        </IconWrapper>
+        <Text size={TextSizes.md}>{label}</Text>
```
andrasbudayssc, line 69: might be worth considering supporting any React element, instead of just strings for the future

andrasbudayssc (approved):
Looks pretty solid to me, special thanks to the default states! 💯

src/components/SortableList/SortableItem.tsx
ajkl2533: I'm counting on that but I wanted to keep it simple for now

src/components/Datatable/Datatable.tsx
ajkl2533: This was already implemented, but I forgot to add it to the prop types and noticed that when I got console warnings:) 
I actually thought that we have onControlChange method but I was wrong :D we probably should add it, but again until I have some clear decision about DS future, til that I consider it in "maintenance" mode and I don't plan to do anything what's not really needed

PR #252, opened 2021-09-01
feat(Banner): add component

PrincAm:
Banner design doesn't follow DS spacing and needs to be updated but UX designer is on PTO. I would like to add Banner in current state to DS not to block depending tickets and then update spacing and other style details when designer will be back. 

From AC
- Each banner should not have more than 2 CTAs and should have adjustable height.
- All CTAs will be classified "Primary"

TODO
- Adjust sizes of layout primitives
- Update documentation
- Discuss other small details like close button hover state color, min width of CTA buttons etc.


src/components/Banner/Banner.tsx
```diff
@@ -0,0 +1,141 @@
...
+import { BannerVariants } from './Banner.enums';
+import { Icon } from '../Icon';
+import { Button } from '../Button';
+import { Inline, Padbox } from '../layout';
+import colors from '../../theme/colors';
+import { IconTypes, SSCIconNames } from '../../theme/icons/icons.enums';
+import { getColor, pxToRem } from '../../utils';
+
+const colorVariants = {
+  [BannerVariants.info]: colors.radiantBlueberry,
```
ajkl2533, line 16: This isn't probably the correct way because if some user updates the theme colors in their application, this component will still use the original one, so I think you should put the color names (not the values) and then use the `getColor`.

src/components/Banner/Banner.tsx
```diff
@@ -0,0 +1,141 @@
...
+  height: 100%;
+  display: flex;
+  align-items: center;
+`;
+
+const StyledIcon = styled(Icon)`
+  color: ${colors.graphite5H};
+`;
+
+const CloseButton = styled.button`
```
ajkl2533, line 50: Could you please add a TODO comment here to take a look at other close buttons we might have in DS to extract that and unify it

src/components/Banner/Banner.tsx
```diff
@@ -0,0 +1,141 @@
...
+`;
+
+const TimesIcon = styled(Icon)`
+  font-size: ${pxToRem(16)};
+`;
+
+const Banner: React.FC<BannerProps> = ({
+  children,
+  variant = BannerVariants.info,
+  cta1,
```
ajkl2533, line 67: I would rather refactor this into an array (let's call it 'actions') since we might need one or three CTA buttons in the future. We could reuse the logic and types (https://github.com/securityscorecard/design-system/blob/alpha/src/components/Datatable/BatchModule/BatchActions/BatchActions.types.ts#L4) from the `BatchModule` in Datatable for this.

ajkl2533 (changes requested):
I have a couple of comments for this PR that needs to be addressed before merging.
Also, I have a feeling that the screenshots are different from the actual design in Figma (there seems to be different font-size for the text and CTA buttons).

src/components/Banner/Banner.tsx
ajkl2533: I see that AC stating there should not be more than two CTAs but the point still stands. Let's make it a `actions` array and we can type it as a tuple and restrict the number of items in the array.

```ts
type MaxLengthArray = readonly [string?, string?];

const a: MaxLengthArray = ['a', 'b']; // <-- this is ok
const b: MaxLengthArray = ['b']; // <-- this is ok
const c: MaxLengthArray = []; // <-- this is ok
const d: MaxLengthArray = ['a', 'b', 'c']; // <-- this will throw: Source has 3 element(s) but target allows only 2.
```

src/components/Banner/Banner.tsx
```diff
@@ -0,0 +1,141 @@
...
+};
+
+const iconVariants = {
+  [BannerVariants.info]: SSCIconNames.infoCircle,
+  [BannerVariants.warn]: SSCIconNames.exclamationTriangle,
+  [BannerVariants.error]: SSCIconNames.ban,
+};
+
+const bgVariants = {
+  [BannerVariants.info]: { r: 235, g: 244, b: 252, a: 0.5 },
```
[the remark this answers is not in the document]
PrincAm: Perfect! Will use it

src/components/Banner/Banner.tsx
PrincAm: Good point, it should be tight to color name

src/components/Banner/Banner.tsx
PrincAm: @ajkl2533 I understand it would be cleaner but AC includes requirement "Each banner should not have more than 2 CTAs" and it makes kind of sense. User can still pass another CTAs within `children` tho. I can at least rename it to `action1` or `firstAction`. WDYT?

ajkl2533: @PrincAm I get it but:
1) with my suggestion we still satisfy the "2 CTAs at most" requirement
2) from my POV whenever you have to index some property you probably doing something wrong
3) when you split it into two props, users could think that each CTA has a different slot (I mean they are rendered in different positions) so they will try to put only `cta2` and that would be kind of strange
4) with the `Action` type we could also support dropdowns 
5) sooner or later there will be a requirement to have 3 CTAs and it's much easier to change the type than change the implementation

PrincAm: You're right, nice clarification. I've used tuple and updated component api.

src/components/Banner/Banner.stories.tsx
```diff
@@ -0,0 +1,128 @@
...
+      {
+        label: 'Action B',
+        name: 'Action B',
+        onClick: action('click-action-b'),
+      },
+    ]}
+    variant={BannerVariants.info}
+    onClose={action('close-banner')}
+  >
+    <Text size="md">
```
ajkl2533, line 57: @PrincAm I'm thinking if we should give people this amount of freedom. If it wouldn't be better to define the text size inside the component itself and then they don't need to think about wrapping children in `Text` component.

src/components/Banner/Banner.tsx
```diff
@@ -0,0 +1,142 @@
...
+  padding: 0;
+`;
+
+const Banner: React.FC<BannerProps> = ({
+  children,
+  variant = BannerVariants.info,
+  actions,
+  onClose,
+}) => (
+  <StyledPadbox variant={variant}>
```
ajkl2533, line 97: Please let's use transient props notation (`$variant={variant}) so we prevent accidental leakage of attributes into DOM.

src/components/Banner/Banner.stories.tsx
PrincAm: Yep, it would be easier for user. I will move `Text` inside

src/components/Banner/Banner.tsx
PrincAm: Sure!

src/components/Banner/Banner.types.ts
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+
+import { BannerVariants } from './Banner.enums';
+import { ActionKinds } from '../../types/action.types';
+
+export type Variants = typeof BannerVariants[keyof typeof BannerVariants];
+
+type MaxLengthArray = readonly [
```
ajkl2533, line 8:
```suggestion
type ActionsArray = readonly [
```

src/theme/colors.enums.ts
```diff
@@ -23,4 +23,6 @@ export const ColorTypes = {
   graphite3H: 'graphite3H',
   graphite4H: 'graphite4H',
   graphite5H: 'graphite5H',
+  pumpkin0: 'pumpkin0',
+  strawberry0: 'strawberry0',
```
ajkl2533, line 27: Could you please group it with their respective colors?
```
pumpkin: ...,
pumpkin0: ...,
strawberry: ...,
strawberry0: ...
```

src/types/customPropTypes.ts
```diff
@@ -0,0 +1,90 @@
...
+  return createChainableTypeChecker(
+    // Validate fn
+    (props, propName, componentName, location, propFullName, secret) => {
+      const value = props[propName];
+      const locationValid = location || 'prop';
+      const propNameValid = propFullName || propName;
+
+      if (isNotArray(value)) {
+        return new CustomPropTypeError(
+          `Invalid ${locationValid} \`${propNameValid}\` supplied to \`${componentName}\`, expected ${types.length}-element array`,
```
ajkl2533, line 65: I think this should say `expected an array with a maximum of two elements` right now it saying that it expects array with strictly given number of elements.

src/components/Banner/Banner.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+const Banner: React.FC<BannerProps> = ({
+  children,
+  variant = BannerVariants.info,
+  actions,
+  onClose,
+}) => (
+  <StyledPadbox $variant={variant}>
+    <Inline align="stretch" stretch="end">
+      <IconPadbox $variant={variant} paddingSize="mdPlus" paddingType="squish">
```
ajkl2533, line 100: Could you please replace props values with respective enums?
`paddingSize` and `gap` - `SpaceSizes`
`paddingType` - `PaddingTypes`
`stretch` - `StretchEnum`
`size` (in <Text /> - `TextSizes`
`variant` (in <StyledButton /> - `ButtonVariants`
```suggestion
      <IconPadbox $variant={variant} paddingSize={SpaceSizes.mdPlus} paddingType={PaddingType.squish}>
```

ajkl2533 (approved):

src/components/Banner/Banner.tsx
PrincAm: Sure, updated

PR #253, opened 2021-09-06
feat(ElementCounter): update styles to remove context text variant

ajkl2533:
In UXDS-88 we are want to remove `context` text variant as it doesn't comply with the accessibility contrast standards. This is the only place in DS where is this variant used. In order to unblock progress in UXDS-88 we changed the appearence of the `ElementCounter` component in `Datatable`.

#### Change
image

Closes UXDS-111


youssefssc (approved):
LGTM

PR #271, opened 2021-09-29
feat(forms): align styles with Figma designs

ajkl2533:
Next stage of styling alignments between DS and Figma designs. This PR solves differences for most of the form fields implemented in DS. We standardized form fields on medium size defined in Figma and the large variants of components will be implemented in future.

⚠️  **DEPRECATION**:
We decided to deprecate `FormFieldWrapper` as is there are currently only 2 occurrences of this component in the platform and can be easily replaced by Layout Primitives.


src/components/forms/SearchBar/SearchBar.tsx
```diff
@@ -83,7 +76,7 @@ const SearchBar: React.FC<SearchBarProps> = ({
   const [isInvalid, setInvalid] = useState<boolean>(false);
   const [query, setQuery] = useState<string>(defaultValue);
   const [typingTimeout, setTypingTimeout] = useState<number>(0);
-  const [isSearching, setIsSearching] = useState<boolean>(false);
+  const [isSearching, setIsSearching] = useState<boolean>(true);
```
youssefssc, line 79: This change results in showing the loading icon by default
Screen Shot 2021-10-01 at 11 22 24 AM

ajkl2533: damn, good catch, I used that for testing since we are missing a bunch of search bar visual tests and forgot to change it back

ajkl2533: @youssefssc Fixed

youssefssc (approved):
great! LGTM

PR #275, opened 2021-10-07
fix(primitives): prevent passing 'undefined' into CSS

ajkl2533:
The `css` function filters out falsy values if is not used `undefined` value is converted to a string and then passed as an invalid value into CSS. The `css` function filters out falsy values.


andrasbudayssc (approved):
Nice one! 💯

PR #281, opened 2021-10-18
fix(Select): fix inconsistencies in design

ajkl2533:
This PR fixes a couple of inconsistencies:
* due to use on border total height was 34px instead of 32px
* multi-value pill remove button had width based on the label length
* clear all indicator was missing hover style
* styles of Filters select box was unified with the Select component
* indicators-related props in theme were generalized for the future use in MultiValueInput


youssefssc (approved):
LGTM

PR #285, opened 2021-10-21
fix(MultiValueInput): fix handlers calling

ajkl2533:
Ref UXD-139


PrincAm (approved):
Good tests coverage 👍

PR #293, opened 2021-11-08
feat(SimpleModal): add SimpleModal component

PrincAm:
PR adds Simple modal component which is the most liberal version of the modal window. Modal content and footer is fully adjustable. 

Screen Shot 2021-11-08 at 17 39 43


src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,153 @@
...
+const Overlay = styled.div`
+  position: fixed;
+  width: 100%;
+  height: 100%;
+  top: 0;
+  background: rgba(0, 0, 0, 0.6);
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  z-index: 1250;
```
ajkl2533, line 37:
```suggestion
  z-index: ${getDepth('modal)};
```

ajkl2533, line 37:
```suggestion
  z-index: ${getDepth('modal');
```

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,153 @@
...
+  border-radius: ${getBorderRadius};
+  background-color: ${getColor('graphite5H')};
+`;
+
+const Title = styled(H3)`
+  margin: 0;
+`;
+
+// TODO Extract close button and unify it across DS
+const CloseButton = styled.button`
```
ajkl2533, line 52: I personally would put some padding in this button so we increase the interactive area so it can be easily used with touchscreens

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,153 @@
...
+  border: none;
+  flex-shrink: 0;
+  cursor: pointer;
+  color: ${getColor('graphite3B')};
+  width: ${pxToRem(16)};
+  height: ${pxToRem(16)};
+`;
+
+const TimesIcon = styled(Icon)`
+  font-size: ${getFontSize('lg')};
```
ajkl2533, line 68: I think you can move this in the `button` component. Icons inherit font-size from parents

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+  return (
+    <Portal>
+      <Overlay>
+        <BaseModal
+          ref={mergeRefs<HTMLDivElement>(onModalRefChange, modalRef)}
+          $width={widthVariants[width]}
+        >
+          <CloseButton aria-label="Close" onClick={onClose}>
+            <TimesIcon name={SSCIconNames.times} type={IconTypes.ssc} />
```
ajkl2533, line 114: `type={IconTypes.ssc}` is default type no need to specify

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,153 @@
...
+  [ModalWidthSizes.sm]: 430,
+  [ModalWidthSizes.md]: 600,
+  [ModalWidthSizes.lg]: 900,
+};
+
+const Overlay = styled.div`
+  position: fixed;
+  width: 100%;
+  height: 100%;
+  top: 0;
```
ajkl2533, line 32: missing `left: 0`

abykovssc:
Noticed strange thing in regression screenshots, modal buttons arranged to the left instead of the right 🤔  f.e. `.storybook/image-snapshots/expected/components_modals_Modal_With Long Content.png` but it might be just a bug from visual regressions

src/components/Modal/hooks/useCalculateContentHeight.ts
```diff
@@ -0,0 +1,49 @@
...
+      setModalHeight(height);
+    }
+  }, []);
+
+  const onFooterRefChange = useCallback((node) => {
+    if (isNotNull(node)) {
+      const { height } = node.getBoundingClientRect();
+      setFooterHeight(height);
+    }
+  }, []);
```
ajkl2533, line 31: I don't think you need `useCallback` here since you are passing only internal variable you are using is `useState` setter which doesn't change during life cycle

src/components/Modal/Modal.tsx
PrincAm: Yes, you're right

src/components/Modal/Modal.tsx
PrincAm: We discussed with Olya the size and it should be at least 24x24. In modal it is fine as the button has absolute position but I can't  extract it and reuse it in Banner, it would break the layout there. We should curate the AC for close button like sizes, color change on hover etc. Anyway, I will increase the padding here

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,153 @@
...
+  overflow-y: auto;
+  max-height: ${({ $maxHeight }) =>
+    isNotNull($maxHeight) ? `${$maxHeight}px` : 'none'};
+`;
+
+const Footer = styled(Padbox)`
+  border-top: 1px solid ${getColor('graphiteHB')};
+`;
+
+const Modal: React.FC<ModalProps> = ({
```
ajkl2533, line 82: It would be great if all new components use forwardRef which is added to the base component, in this case BaseModal.
mergeRefs does support unlimited number of refs

src/components/Modal/hooks/useLockBodyScroll.ts
```diff
@@ -0,0 +1,15 @@
...
+    // Get original body overflow
+    const originalStyle = window.getComputedStyle(document.body).overflow;
+    // Prevent scrolling on mount
+    document.body.style.overflow = 'hidden';
+    // Re-enable scrolling when component unmounts
+    return () => {
+      document.body.style.overflow = originalStyle;
+    };
+  }, []); // Empty array ensures effect is only run on mount and unmount
+};
```
ajkl2533, line 15: please move it to `src/hooks` we will gonna use this also in other components

src/components/Modal/Modal.stories.tsx
```diff
@@ -0,0 +1,129 @@
...
+            tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim
+            veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex
+            ea commodo consequat.
+          </Modal>
+        )}
+      </Inline>
+    </Padbox>
+  );
+};
+
```
ajkl2533, line 69: please disable visual tests for Playground story (it contains only the trigger button :) )
```js
Playground.parameters = {
  screenshot: { skip: true },
};
```

src/components/Modal/Modal.stories.tsx
```diff
@@ -0,0 +1,129 @@
...
+      ...generateControl('select', ModalWidthSizes),
+    },
+    footer: {
+      control: { disable: true },
+    },
+  },
+} as Meta;
+
+const Footer = () => (
+  <Inline gap={SpaceSizes.md} justify="end">
```
ajkl2533, line 28:
```suggestion
  <Inline gap={SpaceSizes.md} justify="flex-end">
```

This will fix footer button alignments in visual tests
cc @abykovssc

src/components/Modal/Modal.tsx
PrincAm: Got it, will bear it in mind

src/components/Modal/hooks/useCalculateContentHeight.ts
PrincAm: Removing this hook as agreed

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,136 @@
...
+  justify-content: center;
+  z-index: ${getDepth('modal')};
+`;
+
+const BaseModal = styled.div<{ $width: number }>`
+  position: relative;
+  display: flex;
+  flex-direction: column;
+  max-height: 66vh;
+  width: ${({ $width }) => pxToRem($width)};
```
ajkl2533, line 52: I missed that spot. We need to have modal horizontally resizable so we need to set width to 100% and given sizes as maximal width
```suggestion
  width: 100%;
  max-width: ${({ $width }) => pxToRem($width)};
```

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,136 @@
...
+    const modalRef = useOuterClick<HTMLDivElement>(onClose);
+
+    useLockBodyScroll();
+
+    return (
+      <Portal>
+        <Overlay>
+          <BaseModal
+            ref={mergeRefs<HTMLDivElement>(modalRef, ref)}
+            $width={widthVariants[width]}
```
ajkl2533, line 105: rename `$width` to `$maxWidth`

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,136 @@
...
+);
+
+export default Modal;
+
+Modal.propTypes = {
+  children: PropTypes.node.isRequired,
+  onClose: PropTypes.func.isRequired,
+  title: PropTypes.string,
+  footer: PropTypes.node,
+  width: PropTypes.oneOf(Object.values(ModalWidthSizes)),
```
ajkl2533, line 135: I would probably change this prop to `size` since you are not passing the width directly.

src/components/Modal/Modal.types.ts
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+import { ModalWidthSizes } from './Modal.enums';
+
+export type Widths = typeof ModalWidthSizes[keyof typeof ModalWidthSizes];
```
ajkl2533, line 5: rename `Widths` to `Sizes`

ajkl2533 (approved):

youssefssc (approved):
LGTM

src/components/Modal/Modal.tsx
PrincAm: Good catch. Fixed 🙂

src/components/Modal/Modal.tsx
PrincAm: Yep. `size` it is

PR #295, opened 2021-11-10
feat(SemanticModal): add SemanticModal component

PrincAm:
This PR is adding Semantic Modal. It is modal for confirmations, alerts and any other proceed steps. It has 4 variants: success, warn, error and ban. Title, main message and at most two actions have to be defined.

Screen Shot 2021-11-10 at 17 36 02


dclerici-ssc (approved):
First touch on this. I had doubt how you pass the actions array as a prop but I found similar code in other components. So I assume it's a standardised way.  Therefore nothing to say.

src/components/SemanticModal/SemanticModal.stories.tsx
```diff
@@ -0,0 +1,155 @@
...
+          ]}
+          message="You can replace this with a local component. You can also just override the text, but keep in mind that you might lose content that way."
+          title="Success"
+          onClose={() => setIsShown(false)}
+          {...args}
+        />
+      )}
+    </Padbox>
+  );
+};
```
ajkl2533, line 68: If you are using `args` you should pass them into the component through the `Playground.args = { ... }` and then you can just use `<SemanticModal {...args} />`  (this is especially valid for the playground story).
You should provide static props only when you want to prevent changing the props by users, but since the published Storybook contains only Docs mode even that is not needed. 
You can take a look at https://github.com/securityscorecard/design-system/blob/alpha/src/components/Dropdown/Dropdown.stories.tsx

src/components/SemanticModal/SemanticModal.stories.tsx
```diff
@@ -0,0 +1,155 @@
...
+        name: 'Primary',
+        onClick: action('click-action-primary'),
+      },
+    ]}
+    message="You can replace this with a local component. You can also just override the text, but keep in mind that you might lose content that way."
+    title="This is a placeholder"
+    variant={SemanticModalVariants.ban}
+    onClose={action('close-modal')}
+  />
+);
```
ajkl2533, line 155: You can replace all this with and it will safe lots of code:
```ts
const SemanticModalTemplate: Story<SemanticModalProps> = (args) => <SemanticModal {...args} />

export const Success = SemanticModalTemplate.bind({});
Success.args = {
  actions: [
      {
        label: 'Secondary',
        name: 'Secondary',
        onClick: action('click-action-secondary'),
      },
      {
        label: 'Primary',
        name: 'Primary',
        onClick: action('click-action-primary'),
      },
  ],
  message: "You can replace this with a local component. You can also just override the text, but keep in mind that you might lose content that way."
  title: "This is a placeholder"
  onClose: action('close-modal'),
  variant: SemanticModalVariants.success
};

export const Warn = SemanticModalTemplate.bind({});
Warn.args = {
  ...Success.args,
  variant: SemanticModalVariants.warn,
};  

export const Error = SemanticModalTemplate.bind({});
Error.args = {
  ...Success.args,
  variant: SemanticModalVariants.error,
};  

export const Ban = SemanticModalTemplate.bind({});
Ban.args = {
  ...Success.args,
  variant: SemanticModalVariants.ban,
};

File: src/components/SemanticModal/SemanticModal.tsx [hunk 3]
```diff
@@ -0,0 +1,120 @@
...
+                      action,
+                      index === 0
+                        ? ButtonEnums.ButtonVariants.outline
+                        : ButtonEnums.ButtonVariants.solid,
+                    ),
+                )}
+              </Inline>
+            </Padbox>
+          </Stack>
+        </Stack>
```
ajkl2533, line 105: This seems to be great ecample when to use `Center` primitive. You can set `isTextCentered` to `true` and get rid of the `CenteredText`. You can also get rid of `justify="center"` on L82 and L87. And add `justify="center"` to L91

ajkl2533 (approved):
Looks good just couple of suggestions 👍

src/components/SemanticModal/SemanticModal.stories.tsx
PrincAm: Good tip. Much shorter

src/components/SemanticModal/SemanticModal.tsx
PrincAm: Updated 😉

PR #296, opened 2021-11-10
feat(Stepper): add Stepper component

ajkl2533:
This PR is adding a Stepper component that allows us to illustrate phased workflows. 

#### Orientation variants
This component is coming in `horizontal` and `vertical` orientation.

**Horizontal**
Screenshot 2021-11-10 at 18 39 53

**Vertical**
Screenshot 2021-11-10 at 18 40 13

#### Scaling
The Stepper component offers in the horizontal variant way how to adapt to the container where it's used. The consumer 
can set `showTextBreakpoint` (default: 600) and if the container shrinks below this size the labels for inactive steps gets hidden.
Screenshot 2021-11-10 at 18 58 30

Ref UXD-224


src/components/Stepper/Step.tsx
```diff
@@ -0,0 +1,156 @@
...
+  max-width: 26ch;
+`;
+
+const StepSummary = styled(Text)`
+  max-width: 30ch;
+`;
+
+const StepButton = styled.button`
+  border: 0 none;
+  background: transparent none;
```
PrincAm, line 26: Shouldn't be just one?

ajkl2533: well yes and no :D it would work in 99 % of cases but :
```
border: 0 === border-width: 0
border: none === border-style: solid
background: transparent === background-color: transparent
background: none === background-image: none
```
TBH inheritance is a bitch 🤣

src/components/Stepper/Step.tsx
```diff
@@ -0,0 +1,156 @@
...
+          )}
+        </div>
+        {!isLast && orientation === StepperOrientations.horizontal && (
+          <StepConnector />
+        )}
+      </>
+    );
+  },
+);
+Step.displayName = 'Step';
```
PrincAm, line 146: 👍 nice, we should use `displayName` more in busy components

ajkl2533: Well display name is usually set for normal React components automatically. In this case the problem was with the test where for some reason wasn't displayName set.
The problem is usually with the styled components that doesn't get the name of the constant

PrincAm (approved):
Awesome! Looking great 🙂

PR #299, opened 2021-11-16
fix(Button): fix Button width

ajkl2533:
Button with set `href` property will stretch accros whole container if
`display` is set to `flex`. We are changing `display` to `inline-flex`
which is more close to actual button implementation as a inline-block element.


andrasbudayssc (approved):
LGTM

PR #302, opened 2021-12-02
feat(forms): extend Checkbox and Radio label support

dclerici-ssc:
Add support for Checkbox and Radio Button label as components.

Closes [UXD-47](https://zitenote.atlassian.net/browse/UXD-47)


src/components/forms/Checkbox/Checkbox.tsx
```diff
@@ -143,7 +143,7 @@ const Checkbox = forwardRef<HTMLInputElement, CheckboxProps>(
     },
     ref,
   ) => {
-    const hasLabel = allPass([isNotUndefined, isNonEmptyString])(label);
+    const hasLabel = allPass([isNotUndefined])(label);
```
ajkl2533, line 146:
```suggestion
    const hasLabel = isNotUndefined(label);
```

src/components/forms/Checkbox/Checkbox.tsx
```diff
@@ -179,7 +179,7 @@ const Checkbox = forwardRef<HTMLInputElement, CheckboxProps>(
 Checkbox.propTypes = {
   name: PropTypes.string.isRequired,
   checkboxId: PropTypes.string,
-  label: PropTypes.string,
+  label: PropTypes.node || PropTypes.string,
```
ajkl2533, line 182: no need to provide both. `PropTypes.node` also covers strings you can take a look into the TS type for that
```
export type ReactNodeLike =
    | {}
    | ReactElementLike
    | ReactNodeArray
    | string
    | number
    | boolean
    | null
    | undefined;

File: src/components/forms/types/forms.types.ts [hunk 3]
```diff
@@ -7,6 +9,6 @@ export interface TogglingProps
   extends TogglingInputProps,
     React.InputHTMLAttributes<HTMLInputElement> {
   name: string;
-  label?: string;
+  label?: React.ReactNode | string;
```
ajkl2533, line 12: same as the prop types `React.ReactNode` is enough :)

ajkl2533:
Please update commit message to 
`feat(forms): extend Checkbox and Radio label support`
`refactor` is meant for changes that doesn't have any impact on functionality and that's not this case, also `refactor` doesn't trigger publishing new package version

src/components/forms/Checkbox/Checkbox.tsx
dclerici-ssc: You're totally right. If we keep both will be redundant.

dclerici-ssc:
Regarding commit messages. I found this [document](https://docs.google.com/document/d/1QrDFcIiPjSLDn3EL15IJygNPiHORgU1_OOAqWjiDU5Y/edit#) in design-system/contributing.md.
I will be studying it

ajkl2533 (approved):

PR #307, opened 2021-12-15
feat(Pill): add Pill component

ajkl2533:
This PR adds a new `Pill` component. This component allows displaying a list of short information (like tags or options). 

#### Sizing
The `Pill` component comes in two sizes (`sm`, `md`) that are the same as the button sizes.
Screenshot 2021-12-15 at 17 44 02

#### Interactions
The `Pill` component support two kinds of interactivity. The `Pill` can be clickable (via using `onClick` property) or if the `onRemove` property is defined it will render the remove button.
Screenshot 2021-12-15 at 17 42 53

#### Adornments
The `Pill` component allows adding custom elements before the component label. This allows us to support a wide range of use-cases as color-coding or additional data to labels.
Screenshot 2021-12-15 at 17 43 47

Ref UXD-303


src/components/Pill/Pill.enums.ts
```diff
@@ -0,0 +1,9 @@
+export const PillVariants = {
+  solid: 'solid',
+  outline: 'outline',
```
abykovssc, line 3:
```suggestion
  outline: 'outline',
  painkillers: 'painkillers'
```
This is a perfect opportunity for an easter egg 😂

ajkl2533: If you come up with proper styling I'm happy to do it :D

PR #314, opened 2022-01-10
feat(Switch): implement Switch (toggle) component

dclerici-ssc:
Implement extend switch component with the variants described [here](https://zitenote.atlassian.net/browse/UXD-305)

Before:
Screen Shot 2022-01-10 at 11 46 31 AM

After:
Screen Shot 2022-01-10 at 11 46 36 AM


src/components/forms/Switch/Switch.stories.tsx
```diff
@@ -1,32 +1,145 @@
...
 export const playground: Story<SwitchProps> = (args) => <Switch {...args} />;
 playground.parameters = {
   chromatic: { disable: true },
 };
-playground.args = { label: 'Public', switchId: 'playground' };
+playground.args = {
+  label: 'Public',
+  switchId: 'playground',
+};
 playground.argTypes = {
```
ajkl2533, line 31: You can remove `argTypes` completely and move type's description to the `Switch.types.ts` file like in. https://github.com/securityscorecard/design-system/blob/alpha/src/components/Pill/Pill.types.ts#L11-L33

src/components/forms/Switch/Switch.stories.tsx
```diff
@@ -1,32 +1,145 @@
...
+
+    <Stack justify="flex-start">
+      <Text>Checked and disabled</Text>
+      <Switch label="True" switchId="withLabel4" defaultChecked isDisabled />
+    </Stack>
+  </Stack>
 );
+
+export const WithFixedWidthAndShortText: Story = () => (
+  <Stack gap="lg">
```
ajkl2533, line 93: You don't need to wrap the component in Stacks when there is only one child in it

src/components/forms/Switch/Switch.stories.tsx
```diff
@@ -1,32 +1,145 @@
...
     description: 'ID to connect input and its label',
   },
+  maxWidth: {
+    description: 'Max width in px',
+  },
 };
 
-export const Default: Story = () => (
-  <Switch label="Private" switchId="default" />
+export const Default: Story = () => <Switch switchId="default" />;
```
ajkl2533, line 43: I miss a story with sizes showcase.

src/components/forms/Switch/Switch.tsx
```diff
@@ -1,70 +1,222 @@
...
+      left: calc(100% - ${pxToRem(2)});
+      transform: translateX(-100%);
+    }
   }
 `;
 
 const Switch: React.FC<
-  SwitchProps & React.PropsWithRef<JSX.IntrinsicElements['input']>
-> = ({ switchId, label, ...props }) => (
+  SwitchProps & React.PropsWithRef<Omit<JSX.IntrinsicElements['input'], 'size'>>
```
dclerici-ssc, line 184: @ajkl2533 regarding your comment on removing argTypes, I tried it before but it didn't work due to this line.

The comments on SwitchProps weren't being read. Apparently is because the React.propsWith.Ref...

I figured out if we add: `<ArgsTable of={Switch} />;` in Switch.stories.tsx, it works. Have you another better solution?

src/components/forms/Switch/Switch.stories.tsx
dclerici-ssc: Answered below. Please check it out.

src/components/forms/Switch/Switch.tsx
ajkl2533: Hm 🤔 not sure if this is correct typing. Why do you use `PropsWithRef`?
I would personally change it to:
```tsx
interface SwitchProps extends Omit<React.HTMLAttributes<HTMLInputElement>, 'size'> {
  ...
}

const Switch = forwardRef<
  HTMLInputElement,
  SwitchProps
>(
({ ... }, ref) => (...)
)
```
That way the descriptions in the `.types` file should be working, I'm using that in Stepper component

src/components/forms/Switch/Switch.stories.tsx
dclerici-ssc: This is working with your suggestion. I will commit changes soon.  I have just copied what you did on Stepper

src/components/forms/Switch/Switch.stories.tsx
dclerici-ssc: Yep. The copy paste feature 👎

src/components/forms/Switch/Switch.stories.tsx
dclerici-ssc: I added a small showcase for each Story 👍

ajkl2533 (approved):
LGTM

PR #315, opened 2022-01-10
refactor(button): change outline button to blue

Mnau-mnau:
Resolves https://zitenote.atlassian.net/browse/UXD-315


src/theme/buttons.ts
```diff
@@ -21,6 +21,16 @@ const buttons: Buttons = {
     },
     outline: {
       primary: {
+        color: colors.blueberryClassic,
+        bgColor: colors.graphite5H,
```
ajkl2533, line 25: According to Figma "no background color in default state" so this should be 'transparent'

ajkl2533 (approved):
LGTM

PR #322, opened 2022-01-21
feat(Pagination): extract Pagination from DataTable component

julesbarr:
Decouple Pagination from Datatable to allow its standalone use.


ajkl2533 (approved):
* There is one place in `GoToPage.types.ts` that needs to update the import of `OnPageChangeFn` type.
* Please update visual tests with `yarn test:storybook:visual:update`.
* Please add export of Pagination component in https://github.com/securityscorecard/design-system/blob/a6a96ab9793a00d78ac183d46f8f491d43396b13/src/components/index.ts#L1

Otherwise looks good. Nice work

PR #332, opened 2022-01-26
build(actions): run unit tests on node@16

ajkl2533:
This PR enables testing on node@16 and also fixes not storing visual test artifacts when the test step fails.


dclerici-ssc (approved):
LGTM

PR #333, opened 2022-01-28
fix(Toast): add primitive layout and apply new spacing rules

dclerici-ssc:
In order to make the design and implementation same, we need to update the Toast component.
The component needs to use Layout primitives for layout and update spacing:

[Closes](https://zitenote.atlassian.net/browse/UXD-358)


src/components/Toast/Toast.tsx
```diff
@@ -51,6 +44,8 @@ const ToastContent = styled(Paragraph)`
   text-decoration: none;
   text-align: left;
 
+  button,
```
ajkl2533, line 47: @dclerici-ssc if you use just `button` as a selector there is no need to also add `button:first/last-child`

src/components/Toast/Toast.tsx
```diff
@@ -79,19 +73,20 @@ const StyledIcon = styled(Icon)`
...
-    <ToastContent as="div" margin="none" size={TextSizes.md}>
-      {children}
-    </ToastContent>
-    <CloseButton aria-label="Close" onClick={onClose}>
-      <StyledIcon name={SSCIconNames.times} />
-    </CloseButton>
+const Toast: React.FC<ToastProps> = ({ onClose, children, width = 400 }) => (
+  <ToastContainer $width={width} onClick={stopPropagation}>
+    <Inline stretch={StretchEnum.start}>
+      <Padbox paddingSize={SpaceSizes.md} paddingType={PaddingTypes.square}>
```
ajkl2533, line 79: `padding-type=square` is the default one so this property can be omitted

src/components/Toast/Toast.tsx
dclerici-ssc: TBH was to solve the problem with buttons within an ` <Inline>` component because inline apply/force margin to children element. As you can see in the TwoActions Toast Story.

So to solve it, I incremented the specificity with &&& button and it works because it overrides the Inline margin.

ajkl2533 (approved):

PR #334, opened 2022-01-28
chore(DataTable): change content alignment

julesbarr:
- change content alignment
- fix visual tests for mac m1


src/components/Datatable/Table/Table.styles.ts
```diff
@@ -46,7 +46,7 @@ export const StyledTable = styled.table`
 
   .ds-table-cell {
     display: flex;
-    align-items: center;
+    align-items: start;
```
ajkl2533, line 49: please replace with `flex-start`

ajkl2533 (approved):

PR #336, opened 2022-01-31
chore(Collapsible): implement layout primitives in Collapsible

julesbarr:


src/components/Collapsible/Collapsible.tsx
```diff
@@ -66,20 +64,24 @@ const Collapsible: React.FC<CollapsibleProps> = ({
...
-        <StyledIcon
-          isRotated={isOpen}
-          name={SSCIconNames.chevronRight}
-          type={IconTypes.ssc}
-        />
-        <div>
-          <Text size={TextSizes.md}>{title}</Text>
-          <Subject>{subject}</Subject>
-        </div>
+      <Header paddingSize="lg" paddingType="squish" onClick={handleHeaderClick}>
```
dclerici-ssc, line 67: Please replace "lg" with `{SpaceSizes.lg}` and "squish" with `{PaddingTypes.squish}`

src/components/Collapsible/Collapsible.tsx
```diff
@@ -66,20 +64,24 @@ const Collapsible: React.FC<CollapsibleProps> = ({
...
+        <Inline align="center" gap="lg">
```
dclerici-ssc, line 68: {SpaceSizes.lg}

src/components/Collapsible/Collapsible.tsx
```diff
@@ -66,20 +64,24 @@ const Collapsible: React.FC<CollapsibleProps> = ({
...
+          <div>
+            <Text size={TextSizes.md}>{title}</Text>
+            <Subject>{subject}</Subject>
+          </div>
+        </Inline>
       </Header>
       {isOpen && (
-        <Content as="div" size={TextSizes.md}>
-          {children}
+        <Content paddingSize="lg" paddingType="square">
```
dclerici-ssc, line 81: Same here: {SpaceSizes.lg}
And remove paddingType="square" because is the default value for Padbox

dclerici-ssc (approved):
LGTM Just some small improvements.

src/components/Collapsible/Collapsible.tsx
ajkl2533: You can remove `margin-right` property in `StyledIcon`

ajkl2533 (approved):

src/components/Collapsible/Collapsible.tsx
julesbarr: is there a reason to use constants as opposed to the string literal?
We seem to be using both variants in the different components

dclerici-ssc: well, It seems to be a way to standardize code. Also, It's intended to be a way to prevent typo errors as well. @ajkl2533 should we run a search and destroy for those literals?

ajkl2533: @dclerici-ssc we can, I hope there are not many strings in the DS tho

PR #339, opened 2022-02-03
feat(theme): implement scalable color taxonomy

ajkl2533:
Closes UXD-38


src/components/Filters/components/Select/styles.ts
```diff
@@ -41,7 +41,7 @@ export const selectStyles: StylesConfig<Option, false> = {
       padding: pxToRem(0, 16),
       background: theme.forms.bgColor,
       border: `${theme.forms.borderWidth} solid ${theme.forms.borderColor}`,
-      borderRadius: theme.borderRadius,
+      borderRadius: theme.radii.default,
```
PrincAm, line 44: TIL radii is plural of radius 😄

PrincAm (approved):
Amazing! I've been waiting for the new taxonomy

src/components/Filters/components/Select/styles.ts
ajkl2533: yes and `indices` is plural for index but that's coming later 🤣

src/utils/helpers.test.ts
```diff
@@ -37,3 +43,40 @@ describe('abbreviateNumber', () => {
     expect(abbreviateNumber(1119)).toBe('1.12K');
   });
 });
+
```
dclerici-ssc, line 46: @ajkl2533 should we also test the new ones?

dclerici-ssc (approved):
LGTM, just a comment on testing getColor function..

src/utils/helpers.test.ts
ajkl2533: We probably can but the question is if it make sense because e.g. `getRadii` is just `path(['radii', type], theme),`.
I added those test to be sure that the changes are backward-compatible

PR #342, opened 2022-02-04
fix(Dropdown): migrate components to use Dropdown implementation

julesbarr:


src/components/ControlDropdown/ControlDropdown.tsx
```diff
@@ -18,19 +17,26 @@ import { Icon } from '../Icon';
...
+import ControlledDropdown from '../Dropdown/ControlledDropdown';
+import { Placements } from '../Dropdown/Dropdown.types';
 
 const CloseButton = styled(Button)`
   padding: 0;
   width: ${pxToRem(16)};
   height: ${pxToRem(16)};
 `;
 
+const placementAdapter: Record<typeof ControlDropdownPlacements, Placements> = {
```
ajkl2533, line 29: I was checking that this component is not used anywhere so feel free to remove the adapter and use the real `*-start` and `*-end` values. We probably can remove the restriction to just bottom position since the popper.js used behind the dropdowns will use this only as prefered default position but will reposition when collision should happen.

src/components/ControlDropdown/ControlDropdown.tsx
```diff
@@ -18,19 +17,26 @@ import { Icon } from '../Icon';
...
+  'bottom-right': 'bottom-end',
+};
+
 const ControlDropdown: React.FC<ControlDropdownProps> = ({
   children,
   title,
   parentRef,
   isOpen = false,
-  onOpen,
   onClose,
```
ajkl2533, line 40: I'm bit worried about removing the `onOpen` prop but it seems to be working. Please remove also a reference to this prop in https://github.com/securityscorecard/design-system/blob/alpha/src/components/Datatable/ControlsModule/ControlsModule.tsx#L221-L226

src/components/Datatable/Dropdown/Dropdown.tsx
```diff
@@ -49,6 +46,15 @@ export const DropdownLink = styled.button<DropdownLinkProps>`
   }
 `;
 
+const placementAdapter = (input: string): Placements => {
```
ajkl2533, line 49: This is also only internal component so we should rather use original values. I would even go further and completely removed the `placement` prop and staticaly set the BaseDropdown `placement` prop to `bottom-end`. And remove prop set from https://github.com/securityscorecard/design-system/blob/alpha/src/components/_internal/BaseTable/columns/actionsColumn.tsx#L70

ajkl2533:
Also please remove `useTooltip` and `useDropdown` custom hooks that are not used anymore.

src/components/ControlDropdown/ControlDropdown.tsx
julesbarr: To me it seem's weird that we have an `onOpen` prop, since is a controlled component, there's no user action that could trigger an `open` action from within the panel

ajkl2533 (approved):
LGTM

PR #343, opened 2022-02-07
feat(SemanticModal): add buttonColor option for primary button

Mnau-mnau:
implements https://zitenote.atlassian.net/browse/UXD-376

### DIFF
--- a/src/components/SemanticModal/SemanticModal.types.ts
+++ b/src/components/SemanticModal/SemanticModal.types.ts
@@ -1,9 +1,11 @@
 import React from 'react';
 
 import { SemanticModalVariants } from './SemanticModal.enums';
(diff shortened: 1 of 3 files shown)


src/components/SemanticModal/SemanticModal.types.ts
```diff
@@ -16,4 +18,5 @@ export interface SemanticModalProps {
   onClose: React.MouseEventHandler;
   actions: ActionsArray;
   variant?: Variants;
+  buttonColor?: ButtonColors;
```
ajkl2533, line 21: let's rename this to `primaryButtonColor` since there can be up to two buttons

ajkl2533 (approved):
LGTM

PR #351, opened 2022-02-14
feat(Toast): add semantic toast variants

dclerici-ssc:
Implement a semantic variant for Toast notifications, same as we have for inline banner.
 
 Also, it fixes warn color for variant 500.

[Closes](https://zitenote.atlassian.net/browse/UXD-150)


src/components/Toast/Toast.tsx
```diff
@@ -68,11 +73,68 @@ const StyledIcon = styled(Icon)`
...
+  [ToastVariants.success]: 16,
+};
+
+const ToastIconPadbox = styled(Padbox)<{ $variant?: ToastProps['variant'] }>`
+  background-color: ${({ $variant }) => getColor(colorVariants[$variant])};
+  display: flex;
+  align-items: center;
+`;
+
+const ToastIconWrapper = styled(FlexContainer)`
```
ajkl2533, line 103: @dclerici-ssc please use normal `div` tag instead of FlexContainer (we are trying to get rid of that thing) and set the flexbox by yourself

ajkl2533: I tried to get rid of this component entirely and it seems that everything works, so we probably can remove this.

ajkl2533 (approved):
Nice job. One comment for consideration

src/components/Toast/Toast.tsx
dclerici-ssc: done!

PR #352, opened 2022-02-14
feat(Banner): add success banner

julesbarr:


src/components/Banner/Banner.tsx
```diff
@@ -94,6 +99,10 @@ const StyledButton = styled(Button)`
   padding: 0;
 `;
 
+const Text = styled(BaseText)`
+  max-width: 75ch;
```
ajkl2533, line 103: `ch` unit is not the same for every font. In our case we probably should use `64ch` to get ~75 characters in one row.

src/components/Banner/Banner.tsx
```diff
@@ -117,27 +126,37 @@ const Banner: React.FC<BannerProps> = ({
...
-                }
-                name={action.name}
-                to={(action as RelativeLinkActionKind<[React.MouseEvent]>).to}
-                variant={ButtonVariants.text}
-                onClick={action.onClick}
-              >
-                {action.label}
-              </StyledButton>
-            ))}
+          <Stack gap={SpaceSizes.sm} justify="start">
```
ajkl2533, line 133: Let's drop the `justify` property, `flex-start` is default value anyway

src/components/Banner/Banner.tsx
```diff
@@ -117,27 +126,37 @@ const Banner: React.FC<BannerProps> = ({
...
+            <Text color={theme.colors.neutral[900]} size={TextSizes.md}>
```
ajkl2533, line 134: `Text` component doesn't have `color` property. But default color value for `Text` component is `neutral.900` so this should be ok to drop.

ajkl2533, line 134: Please use `as="div"`. Default tag for `Text` component is `span` so if somebody put block element inside `children` it would break validity.

ajkl2533 (approved):
Great work. There are three small comments :)

PR #354, opened 2022-02-15
feat(radii): extend radii theme object with common values

dclerici-ssc:
- border radious new values are extracted
- usages of 'getBorderRadius' are replaced with 'getRadii' method
- custom border-radius values are replaced with helper method if applicable

Also, it fixes a border-radius bug on the segmentedToggleItem we had. 

[Closes ](https://zitenote.atlassian.net/browse/UXD-391)


src/components/IconButton/IconButton.tsx
```diff
@@ -4,15 +4,15 @@ import styled from 'styled-components';
...
+import { getFontSize, getRadii, pxToRem } from '../../utils';
 import { Button } from '../Button';
 import { ButtonVariants } from '../Button/Button.enums';
 import { Icon } from '../Icon';
 import { SSCIcons, Types } from '../Icon/Icon.types';
 import { IconButtonProps } from './IconButton.types';
 
 const StyledIconButton = styled(Button)`
-  border-radius: 100%;
+  border-radius: ${getRadii('circle')};
```
julesbarr, line 15: should we create a enum that holds radii values ?

ajkl2533: That's good idea, something similar to what we have for colors

dclerici-ssc: mmm yep.. I was thinking about it when I was doing the replacements. Let's create constants instead

ajkl2533 (approved):
LGTM

PR #355, opened 2022-02-15
feat(Card): add Card component

ajkl2533:


dclerici-ssc (commented):
@ajkl2533 I see subtle difference between storybook and Figma in the horizontal case: 

There is a small space between actions and the outer padding. I had the same problem when I was working on Toast

Screen Shot 2022-02-16 at 9 15 14 AM

Mnau-mnau (approved):

src/components/Card/Card.stories.tsx
```diff
@@ -0,0 +1,185 @@
...
+import { Paragraph } from '../typography';
+import { Stack } from '../layout';
+import Card from './Card';
+import CardHeader from './CardHeader';
+import CardActions from './CardActions';
+import CardContent from './CardContent';
+import CardMedia from './CardMedia';
+import { CardProps } from './Card.types';
+
+const image =
```
dclerici-ssc, line 15: Should we provide a test for another kind of media like video?

dclerici-ssc (commented):
Added small comments..

src/components/Card/Card.stories.tsx
ajkl2533: yeah, I'm not sure how to do it, I don't want to bring external links in the storybook (as it will lead to flaky tests when the origin is not available) and I also don't want to insert videos into the DS repository.
We probably can do a story with external link and disable the test.

ajkl2533:
> @ajkl2533 I see subtle difference between storybook and Figma in the horizontal case:
> 
> There is a small space between actions and the outer padding. I had the same problem when I was working on Toast
> 

@dclerici-ssc is that the Playground story? If yes than this is ok, the problem here is that I didn't wrap the content in Stack with `align="space-between"` if you look for the stories with names With Card Media On Left/Right you can see that the alignment is correct there.
image

src/components/Card/Card.stories.tsx
dclerici-ssc: Well, there are two ways if you don't want to use urls. The most shitty way is to encode into base64 a low-res, low-quality video just to render something. The another one is how I used to encode videos dynamically in the past. Using canvas. Here is an example: https://stackoverflow.com/questions/66592646/can-i-create-a-video-with-javascript

I think it doesn't worth it. A skipped test will be enough if you want to show an example of the power of DS cards :D

dclerici-ssc:
> > @ajkl2533 I see subtle difference between storybook and Figma in the horizontal case:
> > There is a small space between actions and the outer padding. I had the same problem when I was working on Toast
> 
> @dclerici-ssc is that the Playground story? If yes than this is ok, the problem here is that I didn't wrap the content in Stack with `align="space-between"` if you look for the stories with names With Card Media On Left/Right you can see that the alignment is correct there. image

Yes, the story! So no problem

src/components/Card/Card.stories.tsx
ajkl2533: Added showcase of video element

dclerici-ssc (approved):
LGTM

PR #365, opened 2022-02-24
feat(colors): remove legacy colors

Mnau-mnau:
also fixes minor error in colors mapping and updates getColor test


ajkl2533 (approved):
LGTM

PR #376, opened 2022-03-08
feat(Breadcrumb): add breadcrumb component

dclerici-ssc:
This PR adds Breadcrumb component to DS

[Closes](https://zitenote.atlassian.net/browse/UXD-211)


src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,141 @@
...
+        ...(isNotUndefined(breadcrumbItem.props.href) && {
+          href: breadcrumbItem.props.href,
+        }),
+      };
+    }),
+  );
+
+  return (
+    <BreadcrumbsWrapper>
+      <nav aria-label="Breadcrumb">
```
dclerici-ssc, line 122: Just a comment, this doesn't introduce any visual effect. Just for accessibility support: https://www.w3.org/TR/wai-aria-practices/examples/breadcrumb/index.html

 ```
 <nav aria-label="Breadcrumb" class="breadcrumb">
  <ol>
    <li>
      <a href="../../">
        WAI-ARIA Authoring Practices
      </a>
    </li>
    <li>
      <a href="../../#aria_ex">
        Design Patterns
      </a>
    </li>
    <li>
      <a href="../../#breadcrumb">
        Breadcrumb Pattern
      </a>
    </li>
    <li>
      <a href="./index.html" aria-current="page">
        Breadcrumb Example
      </a>
    </li>
  </ol>
</nav>
  ```

src/components/Breadcrumbs/BreadcrumbItem.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+import { BreadcrumbItemProps } from './Breadcrumbs.types';
+import { getColor, getFontWeight } from '../../utils/helpers';
+import { LinkProps } from '../typography/Link/Link.types';
+import { Link } from '../typography/Link';
+
+const BreadcrumbLink = styled(Link).withConfig<
+  LinkProps & { isSelected: boolean }
+>({
+  shouldForwardProp: (property) => !includes(property, ['$isSelected']),
```
ajkl2533, line 14: `$isSelected` is transient property (prefixed with `$`) so this filtering shouldn't be needed

src/components/Breadcrumbs/BreadcrumbItem.tsx
```diff
@@ -0,0 +1,57 @@
...
+})`
+  font-weight: ${getFontWeight('regular')};
+    &,
+    &:not([href]):not([tabindex]) {
+      color: ${({ $isSelected }) =>
+        $isSelected ? getColor('neutral.700') : getColor('neutral.900')}; };
```
ajkl2533, line 20: This one is shorter :)
```suggestion
         getColor($isSelected ? 'neutral.700' : 'neutral.900')};
```

src/components/Breadcrumbs/BreadcrumbItem.tsx
```diff
@@ -0,0 +1,57 @@
...
+      color: ${getColor('primary.700')};
+    }
+`;
+
+const BreadcrumbItem: React.FC<BreadcrumbItemProps> = ({
+  children,
+  isSelected = false,
+  ...props
+}) => {
+  return (
```
ajkl2533, line 39: unnecessary return

src/components/Breadcrumbs/BreadcrumbItem.tsx
```diff
@@ -0,0 +1,57 @@
...
+    <li>
+      <BreadcrumbLink
```
ajkl2533, line 41: We should add handling for the last/current item. Right now if you provide an empty `href` or `to` property, the current item still has on hover effect and is clickable. The last element should be non-interactive. I think we should be safe to check if `href` or `to` are undefined and if yes we should render just `Text` component

src/components/Breadcrumbs/Breadcrumbs.tsx
ajkl2533, line 122: Can we make `BreadcrumbsWrapper` a `nav` element and one DOM node?

src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,141 @@
...
+        <ol>
+          <Inline align="center" gap={SpaceSizes.xs} justify="center">
```
ajkl2533, line 124: You can (and should actually :)) do `<Inline as="ol"...` and save another DOM node. The problem is that `ol` element should contain only `li` elements but right now we are putting `div` inside it.

src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,141 @@
...
+        </DropdownMenu>
+      );
+    };
+
+    return [
+      ...slice(0, itemsBeforeCollapse, allItems),
+      dropdown(),
+      ...slice(allItems.length - itemsAfterCollapse, allItems.length, allItems),
+    ];
+  };
```
ajkl2533, line 88: Can we move this outside of `Breadcrumbs` component it doesn't seem like something that needs to be reinitializes on every render

src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,141 @@
...
+            color={ColorTypes.neutral600}
+            name={SSCIconNames.chevronRight}
+          />
+        </IconWrapper>,
+      );
+    } else {
+      t.push(current);
+    }
+
+    return t;
```
ajkl2533, line 47: Lets do it without mutations
```js
items.reduce((prev, current, index) => {
  if (index < items.length - 1) {
    return [
      ...prev,
      current,
      <IconWrapper key={`separator-${index}`} aria-hidden>
          <Icon
            color={ColorTypes.neutral600}
            name={SSCIconNames.chevronRight}
          />
        </IconWrapper>,
    ] 
  }
  
  return [...prev, current];
}

File: src/components/Breadcrumbs/Breadcrumbs.stories.tsx [hunk 9]
```diff
@@ -0,0 +1,60 @@
...
+import { Meta, Story } from '@storybook/react/types-6-0';
+import { MemoryRouter } from 'react-router-dom';
+
+import { BreadcrumbsProps } from './Breadcrumbs.types';
+import Breadcrumbs from './Breadcrumbs';
+import BreadcrumbItem from './BreadcrumbItem';
+
+export default {
+  title: 'components/Breadcrumbs',
+  component: Breadcrumbs,
```
ajkl2533, line 11: please add `subcomponent` as in https://github.com/securityscorecard/design-system/blob/abb6f70e995aef1953ad6aac4123bf98fde971da/src/components/Stepper/Stepper.stories.tsx#L17

src/components/Breadcrumbs/BreadcrumbItem.tsx
dclerici-ssc: that's right, fixing it

src/components/Breadcrumbs/BreadcrumbItem.tsx
dclerici-ssc: You're right. The last element does not have any hover state.

src/components/Breadcrumbs/Breadcrumbs.tsx
dclerici-ssc: yep. I didn't pick up the habit of using the "as" attribute yet 😞

src/components/Breadcrumbs/BreadcrumbItem.tsx
dclerici-ssc: So by definition, the last item only should be marked as "last item" but we would change to "without href and without to".  That's fine. 
I improved a little bit that part.  Just check it out again please

src/components/Breadcrumbs/Breadcrumbs.test.tsx
```diff
@@ -91,4 +91,28 @@ describe('Breadcrumbs', () => {
     expect(links.findIndex((item) => item.textContent === 'Link1')).toBe(6);
     expect(links.findIndex((item) => item.textContent === 'Link9')).toBe(0);
   });
+
+  it('should be the last breadcrumb item not a link', () => {
```
dclerici-ssc, line 95: I added this test to check the last item. It must not be a link

src/components/Breadcrumbs/BreadcrumbItem.tsx
```diff
@@ -1,34 +1,34 @@
...
-    &:not([href]):not([tabindex]):hover {
-      color: ${getColor('primary.600')};
-    }
-
-    &:active,
-    &:not([href]):not([tabindex]):active {
-      color: ${getColor('primary.700')};
-    }
+  color: ${getColor('neutral.700')};
+`;
```
ajkl2533, line 13: This is unnecessary just use `secondary` variant of Text component:)

ajkl2533 (approved):

PR #378, opened 2022-03-10
feat(Drawer): implement drawer component

Mnau-mnau:
I need to add tests and prepare implementation documentation to show how to handle the drift component visibility.

implements https://zitenote.atlassian.net/browse/UXD-439
https://www.figma.com/file/vnzYfrwIISRYkr2iYE5TcD/branch/DsLAjWuI0ZmY6lsQ65KoMc/DS-Atoms?node-id=15515%3A22020


src/components/Drawer/Drawer.types.ts
```diff
@@ -0,0 +1,16 @@
...
+import { DrawerSizes } from './Drawer.enums';
+
+export type Sizes = typeof DrawerSizes[keyof typeof DrawerSizes];
+
+export interface DrawerProps {
+  children: React.ReactNode;
+  onClose: React.MouseEventHandler;
+  footer?: React.ReactNode;
+  title?: string;
+  size?: string;
```
ajkl2533, line 12:
```suggestion
  size?: Sizes;
```

src/components/Drawer/Drawer.stories.tsx
```diff
@@ -0,0 +1,176 @@
...
+);
+
+export const WithOverflowingTitle: Story = () => (
+  <Drawer
+    footer={<Footer />}
+    size="md"
+    title="This is a title that has no end, therefore it goes on and on until it fills all the space of its container and wants to go even further."
+    onClose={action('close-modal')}
+  >
+    <H2 style={{ margin: 0 }}>This is a placeholder</H2>
```
ajkl2533, line 114: We don't want to promote styling through `style` property much. You can wrap it in `Stack` and it should work the same way

src/components/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,178 @@
...
+      containerId: portalsContainerId,
+      internalShowHide: false,
+      autoRemoveContainer: false,
+    });
+    const hasFooter = isNotUndefined(footer);
+    const hasAdornment = isNotUndefined(adornment);
+
+    const drawerRef = useOuterClick<HTMLDivElement>(onClose);
+
+    useLockBodyScroll();
```
ajkl2533, line 117: Be careful, we should be locking body only when there is overlay. If there is no backdrop, the body should be still interactive.

src/components/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+const Drawer = forwardRef<HTMLDivElement, DrawerProps>(
+  (
+    {
+      title,
+      onClose,
+      footer,
+      size = DrawerSizes.md,
+      adornment,
+      hasNoBackdrop,
```
ajkl2533, line 100: I would rather use positive property `hasNoBackdrop` -> `hasBackdrop=true`

src/components/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,178 @@
...
+  top: 0;
+  left: 0;
+  background: rgba(0, 0, 0, 0.25);
+  display: flex;
+  align-items: flex-start;
+  justify-content: flex-end;
+  z-index: ${getDepth('modal')};
+`;
+
+const NoOverlay = styled.div`
```
ajkl2533, line 40: This component should not exist, we need interactive body when when overlay is disabled

src/components/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,178 @@
...
+        ref={mergeRefs<HTMLDivElement>(drawerRef, ref)}
+        $maxWidth={widthVariants[size]}
+        data-testid={dataTestId}
+      >
+        <Header>
+          <Inline stretch={StretchEnum.start}>
+            <Padbox paddingSize={SpaceSizes.lg} paddingType="squish">
+              <Inline align="center" gap={SpaceSizes.md}>
+                {hasAdornment && <Adornment>{adornment}</Adornment>}
+                {isNotUndefined(title) && (
```
ajkl2533, line 129: Title should be required property, so no need to check if it's undefined

src/components/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,178 @@
...
+import { StretchEnum } from '../layout/Inline/Inline.enums';
+
+const widthVariants = {
+  [DrawerSizes.sm]: 360,
+  [DrawerSizes.md]: 480,
+  [DrawerSizes.lg]: 720,
+  [DrawerSizes.xl]: 960,
+};
+
+const Overlay = styled.div`
```
ajkl2533, line 27: There is very similar `Overlay` component in `Modal` we should extract it into `src/_internal/BaseOverlay` and reuse it for both components

src/components/Drawer/Drawer.test.tsx
```diff
@@ -0,0 +1,115 @@
...
+        footer={<Footer />}
+        dataTestId="drawer"
+      >
+        <Stack>
+          <Paragraph>Whatever happens, happens here</Paragraph>
+        </Stack>
+      </Drawer>,
+    );
+
+    const footer = screen.getByTestId('footer');
```
ajkl2533, line 39: You can actually test not the footer but the button it contains like:
```js
const footerButton = screen.getByRole('button',  { name: /Submit/i };
expect(footerButton).toBeInDocument();
```
https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-the-wrong-query

src/components/Drawer/Drawer.stories.tsx
ajkl2533: We probably can extract the Drawer content into a separate component like the `Footer`. So we don't need to repeat the same code in every story

src/components/Drawer/Drawer.tsx
Mnau-mnau: overlay only takes the width of the modal, so it doesn't go over the rest of the page and holds z-index property

Mnau-mnau: renamed to Drawer wrapper

PR #384, opened 2022-03-14
fix(Breadcrumbs): fix order and options showed in dropdown

dclerici-ssc:
We decided to change the order of the elements in breadcrumb overflow-dropdown not to be in reverse order. Also, dropdown options list only show the hidden elements, not all breadcrumb items. 

[Closes](https://zitenote.atlassian.net/browse/UXD-211)

### DIFF
--- a/src/components/Breadcrumbs/Breadcrumbs.tsx
+++ b/src/components/Breadcrumbs/Breadcrumbs.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import PropTypes from 'prop-types';
-import { reverse, slice } from 'ramda';
+import { slice } from 'ramda';
 import styled from 'styled-components';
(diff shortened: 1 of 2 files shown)


src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -98,8 +98,11 @@ const Breadcrumbs: React.FC<BreadcrumbsProps> = ({ children, ...props }) => {
     });
   });
 
-  const allDropdownActions = reverse(
-    React.Children.map(children, (breadcrumbItem) => {
+  const allDropdownActions = pipe(
```
ajkl2533, line 101: you can use `slice(1, -2)` instead of dropping

ajkl2533 (approved):
LGTM, one small improvement

PR #400, opened 2022-03-17
feat(DatePicker): implement year picker

julesbarr:


src/components/SingleDatePicker/SingleDatePicker.stories.tsx
```diff
@@ -46,3 +46,30 @@ export const WithCustomPlaceholder: Story = () => {
...
 };
+
+export const OpenDatepicker: Story = () => {
+  const [dateValue, handleChangeDate] = useState<Date>(null);
+
+  return (
+    <SingleDatePicker
+      placeholder="Custom placeholder"
+      value={dateValue}
+      isOpen
```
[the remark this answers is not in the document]
ajkl2533: I think this is ok, we expose that also for other components, but I would change it to:
* `defaultIsOpen`
* `defaultIsYearPickerOpen`
to remain consistent with other components

src/components/_internal/BaseSingleDatePicker/CustomHeader.tsx
```diff
@@ -0,0 +1,86 @@
...
+`;
+
+export const DatePickerCustomHeader: React.FC<{
+  decreaseMonth: () => void;
+  increaseMonth: () => void;
+  toggleYearPicker: () => void;
+  date: Date;
+}> = ({ decreaseMonth, toggleYearPicker, date, increaseMonth }) => {
+  return (
+    <Inline justify="space-between">
```
ajkl2533, line 72:
```suggestion
    <Inline gap={SpaceSizes.sm} stretch={2}>
```

src/components/_internal/BaseSingleDatePicker/CustomHeader.tsx
```diff
@@ -0,0 +1,86 @@
...
+      <MonthSelector type="button" onClick={decreaseMonth}>
```
ajkl2533, line 73: It would be great if we can list through years in the year picker mode. There are props `increaseYear` and `decreaseYear` provided that we can use for this.
```suggestion
      <MonthSelector type="button" onClick={isYearPickerOpen ? decreaseYear : decreaseMonth}>
```

ajkl2533 (commented):
Nice work, but I want to suggest couple more improvements:
1) When a year is selected date resets to Jan 1 of the selected year. It would be nice if we can change only the year and leave rest of the previously selected date untouched. Like>
March 23 2022 -> select year 2020 -> March 23 2020 (currently Jan 1 2020).
2) can you enable this year picker also for the DateRangePicker

src/components/_internal/BaseSingleDatePicker/CustomHeader.tsx
```diff
@@ -0,0 +1,86 @@
...
+        <ChangeMonthIcon direction="left" />
+      </MonthSelector>
+      <MonthIndicator onClick={() => toggleYearPicker()}>
+        <Padbox paddingSize={SpaceSizes.md} paddingType={PaddingTypes.squish}>
+          {`${getMonth(date)} ${getYear(date)}`}
```
ajkl2533, line 78: The style of the `{month} {year}` title is different from Figma design

src/components/SingleDatePicker/SingleDatePicker.stories.tsx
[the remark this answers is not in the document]
julesbarr: I think we don't actually need to expose this, but is a nice API addition, and it also helps to have more meaningful visual regression tests, otherwise we're just testing the text input

src/components/_internal/BaseDateRangePicker/BaseDateRangePicker.tsx
```diff
@@ -1,21 +1,28 @@
...
 import {
   BaseDateRangePickerPropTypes,
   BaseDateRangePickerProps,
   BaseDateRangePlaceholderPropTypes,
 } from './BaseDateRangePicker.types';
+import { StyledDatePicker as SingleStyledDatePicker } from '../BaseSingleDatePicker/SingleDatePicker';
+import { BaseSingleDatePicker } from '../BaseSingleDatePicker';
+import { pxToRem } from '../../../utils';
 
 const StyledDatePicker = styled(FlexContainer)`
```
ajkl2533, line 18: We can change this into:
```js
const StyledDatePicker = styled.div`
  ${dateRangePickerStyles}
  width: 100%;
`;
```
and then wrap both `BaseSingleDatePicker` in the `Inline` like:
```jsx
<StyledDatePicker>
  <Inline gap={SpaceSizes.sm} stretch={InlineEnums.StretchEnum.all}>
     <BaseSingleDatePicker ... />
     <BaseSingleDatePicker ... />

ajkl2533 (approved):
Very nice work 👍 I have one last suggestion to remove FlexContainer

PR #404, opened 2022-03-22
fix(Breadcrumbs): remove circular dependency issue

dclerici-ssc:
This PR adds the Breadcrumb component export. Also, I've addressed a circular dependency problem that fortunately helps fixing an alignment problem for the dropdown ellipsis icon.


ajkl2533 (approved):
LGTM, but I would change commit message to `fix(Breadcrumbs): remove circular dependency issue`. Since that is the bigger part of this PR.

PR #407, opened 2022-03-23
feat(UserAvatar): add new UserAvatar component

dclerici-ssc:
This PR adds a new UserAvatar component.

[Closes](https://zitenote.atlassian.net/browse/UXD-461)


src/components/UserAvatar/UserAvatar.tsx
```diff
@@ -0,0 +1,59 @@
...
+  size = UserAvatarSizes.md,
+  ...props
+}) => (
+  <StyledButton
+    color={ButtonColors.primary}
+    size={size}
+    variant={ButtonVariants.solid}
+    {...props}
+  >
+    {normalizeString(children.substring(0, 2))}
```
ajkl2533, line 50: Hm I think we should trim the string first and then take the substring, maybe we can change `normalizeString` into:
```js
const normalizeString = pipe(
  trim,
  take,
  toUpper
)

...

{normalizeString(children)}

File: src/components/UserAvatar/UserAvatar.tsx [hunk 2]
```diff
@@ -0,0 +1,59 @@
...
+`;
+const normalizeString = pipe(trim, toUpper);
+
+const UserAvatar: React.FC<UserAvatarProps> = ({
+  children,
+  size = UserAvatarSizes.md,
+  ...props
+}) => (
+  <StyledButton
+    color={ButtonColors.primary}
```
ajkl2533, line 45: We should make this with custom (branding) colors, we will change this later in Q3 when wi will change button colors. We will use this in the header and if we make this blue it can look strange in the context of our logo.
For now let's use arbitrary colors and leave TODO comment here to migrate when we change button colors.
```
default = #6641F3
hover = #5737CF
active = #482EAD

File: src/components/UserAvatar/UserAvatar.tsx [hunk 3]
```diff
@@ -0,0 +1,59 @@
...
+  [UserAvatarSizes.sm]: UserAvatarSmall,
+  [UserAvatarSizes.md]: UserAvatarMedium,
+};
+
+const StyledButton = styled(Button)<{ size: Sizes }>`
+  border-radius: ${getRadii('round')};
+  font-size: ${getFontSize('md')};
+  font-weight: ${getFontWeight('regular')};
+  line-height: ${getLineHeight('md')};
+  ${({ size }) => userAvatarSizes[size]};
```
ajkl2533, line 35: We should set width rather than manipulate with padding, we have button heights defined in the theme so we can use those to make it perfect circle.
```suggestion
  width: ${({ size, theme }) =>
    pipe(getButtonHeight(size), pxToRem)({ theme })};
```

src/components/UserAvatar/UserAvatar.tsx
```diff
@@ -0,0 +1,59 @@
...
+  pxToRem,
+} from '../../utils/helpers';
+import { UserAvatarSizes } from './UserAvatar.enums';
+
+const UserAvatarSmall = css`
+  padding: ${pxToRem(4, 2.5)};
+`;
+const UserAvatarMedium = css`
+  padding: ${pxToRem(8, 6.5)};
+`;
```
ajkl2533, line 23: Remove

src/components/UserAvatar/UserAvatar.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+const userAvatarSizes = {
+  [UserAvatarSizes.sm]: UserAvatarSmall,
+  [UserAvatarSizes.md]: UserAvatarMedium,
+};
+
+const StyledButton = styled(Button)<{ size: Sizes }>`
```
ajkl2533, line 30: I'm thinking if we should support a non-button variant (as `div` element, without hover, focus, active states)

src/components/UserAvatar/UserAvatar.types.ts
```diff
@@ -0,0 +1,15 @@
+import { UserAvatarSizes } from './UserAvatar.enums';
+
+export type Sizes = typeof UserAvatarSizes[keyof typeof UserAvatarSizes];
+
+export interface UserAvatarProps {
+  /**
+   * Used to render text elements inside the Avatar.
+   * It Will render the first two characters of the string always.
+   */
+  children: string;
```
ajkl2533, line 10: I'm thinking if we should use rather `label` property instead of `children` because if children are used people tend to customize the content and ask why we are restricting just to string

src/components/UserAvatar/UserAvatar.tsx
dclerici-ssc: Nice, I didn't realize that take will work with Strings also.

src/components/UserAvatar/UserAvatar.tsx
dclerici-ssc: Just to be sure that you have read the Figma message :D 

@ajkl2533 I went with the primary due to this comment on FIgma: 
https://www.figma.com/file/vnzYfrwIISRYkr2iYE5TcD/DS-Atoms?node-id=17126%3A21530

Known issues:
"the brand color is not supported on interactive elements yet, so we are using the primary colors as a fallback"

So, would you like to go with the new brand ones instead?

src/components/UserAvatar/UserAvatar.types.ts
dclerici-ssc: @ajkl2533 I think you are right. Because even Pill and Badge are not using children and they could use it. So, I will change it to the label to be more aligned with overall

src/components/UserAvatar/UserAvatar.tsx
dclerici-ssc: Hmm well the avatar component on MUI is non-interactive by default.  

Can we use an "interactive=false" attribute? By default true? Not sure if there is a use case in which would be used.

src/components/UserAvatar/UserAvatar.tsx
ajkl2533: @dclerici-ssc I'm aware of that comment. But even with that default state is in Brand color. I think we should go with the purple color it seems like the right solution, even if we need to refactor it later.

src/components/UserAvatar/UserAvatar.tsx
ajkl2533: We can leave it for now, but I expect that we will get a request to show this avatar e.g. in setting in the user profile. But we can easily add this later

ajkl2533 (approved):
LGTM

PR #409, opened 2022-03-24
fix(UXD-505): Fix button component 'as' prop

marekpukaj:
[UXD-505](https://zitenote.atlassian.net/browse/UXD-505)


ajkl2533 (approved):
LGTM

PR #410, opened 2022-03-24
chore(dependencies): Upgrade dependencies to closest minor version

Mnau-mnau:
implements [UXD-485](https://zitenote.atlassian.net/browse/UXD-485)


package.json
```diff
@@ -44,29 +44,29 @@
     "@dnd-kit/sortable": "^2.0.1",
     "@dnd-kit/utilities": "^1.0.3",
     "@fluentui/react-component-ref": "^0.59.0",
-    "@fortawesome/fontawesome-svg-core": "^1.2.35",
-    "@fortawesome/react-fontawesome": "^0.1.14",
-    "@popperjs/core": "^2.10.2",
+    "@fortawesome/fontawesome-svg-core": "1.3.0",
```
ajkl2533, line 47: Be careful we should leave the caret at the beginning like `^1.3.0` without caret we would fix the version

ajkl2533 (approved):
LGTM

PR #422, opened 2022-04-01
chore(dependencies): upgrade TS from 3.9.10 to 4.6.2

dclerici-ssc:
This PR upgrades Typescript from 3.9.10 to 4.6.23

There are a few comments below, regarding some changes that were needed to compile.

- tslib latest version was added as a dependency. This is because we are using es5 as a target for this project (tsconfig.json). We can remove this dependency if we decide to move to the es6 target. 
- There were updated some files with '../..' import, to remove TS warnings running yarn build.
- typescript-eslint was upgraded to 5.17.0 due to the upgrade of TS. This library will need node LTS versions installed. Tested with v16.14.2 

[Closes](https://zitenote.atlassian.net/browse/UXD-488)


ajkl2533:
LGTM, but I would like to do two things here:
1) we should add `.nvmrc` file with the required node version, so any contributor knows what they need to use
2) since we'll need to use a different node version than the platform I would like to extend `build` step in CI with node version matrix and try to build the package with versions 12, 14, 16 (we can use the same setup as we are using for unit tests)

.github/workflows/CD.yaml
```diff
@@ -250,14 +250,17 @@ jobs:
     if: "github.event_name == 'push' && (github.ref == 'refs/heads/main' || github.ref == 'refs/heads/alpha') && !contains(github.event.head_commit.message, 'chore(release):')"
     needs: [unit-tests, visual-tests, types, lint, lint-css, betterer, build]
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node: ['12', '14', '16']
```
ajkl2533, line 255: This is not the place though :) the correct step start on line 210. This would cause error in better case and three releases in worse case

PR #423, opened 2022-04-04
chore(bundle): add terser plugin to reduce bundle size

dclerici-ssc:
This PR adds [npm: rollup-plugin-terser](https://www.npmjs.com/package/rollup-plugin-terser) to reduce bundle size. 

We are using default parameters. 

Tested locally, build/index.es.js seems to reduce almost to half. 

Also, the warning from ssc-ui-core goes away

Before:
Screen Shot 2022-04-04 at 2 48 39 PM

After:
Screen Shot 2022-04-04 at 2 48 31 PM

[Closes](https://zitenote.atlassian.net/browse/UXD-474)

### DIFF
--- a/package.json
+++ b/package.json
@@ -136,6 +136,7 @@
     "rollup": "^2.45.0",
     "rollup-plugin-auto-external": "^2.0.0",
     "rollup-plugin-peer-deps-external": "^2.2.2",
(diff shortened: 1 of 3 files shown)


ajkl2533 (approved):
LGTM

PR #426, opened 2022-04-11
chore(dependencies): update husky

Mnau-mnau:
implements https://zitenote.atlassian.net/browse/UXD-495


.husky.json
```diff
@@ -0,0 +1,6 @@
+{
```
ajkl2533, line 1: We should remove this file. In v7 there should be used the separate hooks instead of this config file

PR #431, opened 2022-04-12
chore(package): upgrade semantic-release ecosystem

ajkl2533:
I did a couple of fixes in the CI to make dependencies and build caches dependent on the node version. I also fixed the issue in the `build` step where CI was using the build from the cache instead of testing if the package can be built correctly.

Since `semantic-release` can be only run on node@14.17+ I had to move it to `optionalDependencies` as it was failing the install on `unit-test` and `build` steps on node@12. With this change, we are saying we are fine if those packages aren't installed since we need them only in the `build-and-publish` step.

Closes UXD-494

### DIFF
--- a/.github/workflows/CD.yaml
+++ b/.github/workflows/CD.yaml
@@ -22,12 +22,13 @@ jobs:
           node-version: ${{ matrix.node }}
 
       - name: Cache node modules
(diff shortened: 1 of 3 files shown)


dclerici-ssc (approved):
LGTM

PR #437, opened 2022-04-14
docs(Drawer): add subheader example

ajkl2533:
Ref UXD-554


src/components/Drawer/Drawer.stories.tsx
```diff
@@ -204,3 +206,94 @@ export const WithSemanticModal: Story = () => {
...
+          </Paragraph>
+        ))}
+      </Stack>
+    </Stack>
+  </Drawer>
+);
+
+const codeExample = `
+const Subheader = styled(Padbox)\`
+  /* make header stick bellow the header */
```
dclerici-ssc, line 266: I think there is a typo here: should be "make **subheader** stick below the header"

PR #443, opened 2022-04-19
chore(deps): upgrade type definitons to latest versions

dclerici-ssc:
This PR upgrades type definitions to latest versions. 

@types/history                             latest  4.7.8     ❯  5.0.0          
@types/jest                                latest  26.0.22   ❯  27.4.1         
@types/node                                latest  14.14.37  ❯  16.11.27 (latest version for Node v16)        
@types/react-select                        latest  3.1.2     ❯  4.0.18 (latest version for react-select v4)

[Closes](https://zitenote.atlassian.net/browse/UXD-497)


ajkl2533:
@dclerici-ssc it's not really true that we don't need `@types/react-select` that would be the case if we use `react-select@5` (see https://github.com/JedWatson/react-select/releases/tag/react-select%405.0.0). But the problem is we cannot use v5 since it doesn't support IE11, which is supported by platform, so we cannot upgrade until we drop the IE support. So we still should upgrade types I saw that there were some types test failing, I'm not sure how versioning works for the `@types` packages maybe we can upgrade to the latest version for v4 types

dclerici-ssc:
> @dclerici-ssc it's not really true that we don't need `@types/react-select` that would be the case if we use `react-select@5` (see https://github.com/JedWatson/react-select/releases/tag/react-select%405.0.0). But the problem is we cannot use v5 since it doesn't support IE11, which is supported by platform, so we cannot upgrade until we drop the IE support. So we still should upgrade types I saw that there were some types test failing, I'm not sure how versioning works for the `@types` packages maybe we can upgrade to the latest version for v4 types

You are right. I misunderstood the types-thing upgrading to v4. I thought it was since v4. I didn't see the upgrade to v5 release guide because I wasn't upgrading react-select to v5 :D 
So, I [found](https://github.com/JedWatson/react-select/discussions/4927#discussioncomment-1681060) that latest version for types-v4 is 4.0.18

PR #451, opened 2022-04-25
fix(Tabs): change sm tab font size to h5

julesbarr:


src/components/_internal/BaseTabs/BaseTabLabel.tsx
```diff
@@ -42,6 +42,11 @@ const smallUnderlineSize = css`
   line-height: ${getLineHeight('lg')};
 `;
 
+const extraSmallUnderlineSize = css`
+  font-size: ${getFontSize('h5')};
+  line-height: ${getLineHeight('lg')};
+`;
+
```
ajkl2533, line 49: @julesbarr The plan was not to add a new variant but to update `sm` size. Please remove this `xs` size and change `sm` to use `h5` font-size. 
Since `h4` and `h5` are currently the same it won't affect current usages in the platform but when we do the font change we also change the `h5` size in the new theme and that will cause desired change.

ajkl2533 (commented):
LGTM, just remove the `xs` and update `sm` to use `h5` size

ajkl2533 (approved):
Nice 👍

PR #459, opened 2022-05-02
chore(package): update visual-regressions deps

ajkl2533:


Mnau-mnau (approved):
works on M1 it seems, so 👍

PR #460, opened 2022-05-02
chore(package): upgrade Storybook to v6.4

ajkl2533:


dclerici-ssc (approved):
LGTM. I've just tested locally and it works

PR #463, opened 2022-05-02
feat(DataTable): expose pageButtonsCount config

julesbarr:


dclerici-ssc (commented):
Hey @julesbarr looks really good. One small comment, would be nice to have an example (story) using that pageButtonsCount prop because is a feature requested by consumer. So, we can refer to it when we communicate

julesbarr:
@dclerici-ssc I added the prop to the story controls so it's configurable now

dclerici-ssc (approved):
LGTM. Please don't merge. We need to upload this along with the pagination fix.

src/components/Datatable/Datatable.tsx
```diff
@@ -49,6 +49,7 @@ function Datatable<D extends Record<string, unknown>>({
   isBatchModuleEnabled = true,
   controlsConfig = {},
   tableConfig = {},
+  pageButtonsCount,
```
ajkl2533, line 52: This should be hidden inside the `tableConfig` property as written in the ticket. This setting is coupled only with the table itself and there is no need to expose it as a separate property

ajkl2533 (commented):
Looks good, we just need to move the property inside `tableConfig`

PR #467, opened 2022-05-04
fix(Tabs): change hover and selected colors

dclerici-ssc:
This PR fixes color for hover and selected state for Tabs. 

[Closes](https://zitenote.atlassian.net/browse/UXD-552)

### DIFF
--- a/src/components/_internal/BaseTabs/BaseTabLabel.tsx
+++ b/src/components/_internal/BaseTabs/BaseTabLabel.tsx
@@ -74,15 +74,26 @@ const segmentedSizes = {
   [BaseTabSizes.md]: mediumSegmentedSize,
 };
 
+const commonHoverTab = css<BaseLabelProps>`
+  ${({ $isSelected, $color, theme }) =>
+    $color && $isSelected
+      ? lighten(0.1, getColor($color, { theme }))
+      : getColor('primary.600')}
+`;
+
 const underlineTab = css<BaseLabelProps & { size: Sizes }>`
   ${({ size }) => underlineSizes[size] || underlineSizes[BaseTabSizes.md]};
   padding-bottom: ${getSpace(SpaceSizes.xxs)};
   border-bottom: 2px solid
     ${({ $isSelected, $color }) =>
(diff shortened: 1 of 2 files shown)


julesbarr (approved):
LGTM

src/components/Tabs/Tabs.stories.tsx
```diff
@@ -78,7 +78,7 @@ Playground.argTypes = {
 Playground.args = { selectedValue: 'inventory' };
 
 export const OnlyTabs: Story = () => (
-  <Tabs selectedValue="overview" onSelectTab={action('Select Tab')}>
+  <Tabs selectedValue="inventory" onSelectTab={action('Select Tab')}>
```
ajkl2533, line 81: @dclerici-ssc why did you changed this? In this case should be active the second tab with the custom color as a example of the customization

src/components/_internal/BaseTabs/BaseTabLabel.tsx
```diff
@@ -74,15 +74,26 @@ const segmentedSizes = {
...
   ${({ size }) => underlineSizes[size] || underlineSizes[BaseTabSizes.md]};
   padding-bottom: ${getSpace(SpaceSizes.xxs)};
   border-bottom: 2px solid
     ${({ $isSelected, $color }) =>
-      $isSelected ? getColor($color) : getColor('neutral.500')};
+      $isSelected
+        ? getColor('primary.500')
+        : $color
+        ? getColor($color)
+        : getColor('neutral.500')};
```
ajkl2533, line 93: I don't think this is correct. It should work that the `$color` is the selected color, if `$color` is not provided then we use the default primary. It's not in Figma but this will break the IHS report modal where we are using their brand color for this special report

dclerici-ssc: oh got it. I thought the custom color was for the normal case. And that's the reason why I've changed the test. Fixing it

ajkl2533 (approved):
LGTM

PR #468, opened 2022-05-04
feat(Datatable): Add hide dropdown on action item click

julesbarr:
Note: this needs to be rebased once the buttonsPageCount changes are merged, so please ignore those

### DIFF
--- a/src/components/Dropdown/ControlledDropdown.stories.tsx
+++ b/src/components/Dropdown/ControlledDropdown.stories.tsx
@@ -61,6 +61,7 @@ export const UsingWithAnotherComponentControlled: Story<DropdownProps> = (
       </span>
       <ControlledDropdown
         {...args}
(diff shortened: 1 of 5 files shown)


src/components/Datatable/Datatable.tsx
```diff
@@ -49,7 +49,11 @@ function Datatable<D extends Record<string, unknown>>({
   isBatchModuleEnabled = true,
   controlsConfig = {},
   tableConfig = {},
-}: DatatableProps<D>): React.ReactElement {
+  pageButtonsCount,
+  shouldCloseActionsDropdownOnItemClick,
```
ajkl2533, line 53: @julesbarr I think we can actually drop this property and close actions whenever the action is clicked. I don't see many reasons why anyone would want to keep it open after click and dismissing the menu is default action anyway on click

PR #470, opened 2022-05-06
fix: change npm registry

julesbarr:


dclerici-ssc (approved):
LGTM. Let's try if it's enough

PR #477, opened 2022-05-10
build: remove Node@12 tests

ajkl2533:
* replaces Node@12 with Node@18
* upgrades `actions/setup-node` to latest version
* fix some typing inconsistencies
* add uploading bundle stats during the build step
* change handling of visual regression artifacts (tar before upload and upload only when test fails)
* move semantic release back to dev deps to prevent leaking in the consumer's yarn.lock (this was done because installing was filing when used with Node@12)


dclerici-ssc (approved):
LGTM :)

PR #481, opened 2022-05-17
feat(colors): change semantic color values

Mnau-mnau:
implements https://zitenote.atlassian.net/browse/UXD-600


ajkl2533:
We should change text color in the Badge component to `neutral.1000` for **warning** and **success** variant to comply with WCAG color contrast rules

Mnau-mnau:
> We should change text color in the Badge component to `neutral.1000` for **warning** and **success** variant to comply with WCAG color contrast rules

that's part of https://zitenote.atlassian.net/browse/UXD-531 I'd say

PR #483, opened 2022-05-17
feat(CloseButton): add inverted color option

Mnau-mnau:
implements https://zitenote.atlassian.net/browse/UXD-601
it changes the default color of the close button to pure black, the inverted version is white


ajkl2533 (approved):
One small comment

PR #490, opened 2022-05-23
feat(Banner): add new style to banner

Mnau-mnau:
partially implements https://zitenote.atlassian.net/browse/UXD-531, behaviour needs to be researched in related spike as to how we will use it in platform later


ajkl2533:
Well postpone merging of this until we have decision how to handle stacking

julesbarr (approved):

src/theme/icons/errorCircle.ts
```diff
@@ -0,0 +1,17 @@
+import {
```
ajkl2533, line 1: Is this from FontAwesome? If yes we are missing license comment here

ajkl2533 (approved):
LGTM

PR #496, opened 2022-05-31
refactor(button): remove the size property altogether

dclerici-ssc:
This PR is the latest part of the "retirement of button size". 

We agreed to retire the large and small button size options going forward and replace them both with the Medium size.

[Closes](https://zitenote.atlassian.net/browse/UXD-586)


ajkl2533 (approved):
LGTM

PR #497, opened 2022-06-01
feat: add Callout component

julesbarr:


src/components/Callout/Callout.stories.tsx
```diff
@@ -0,0 +1,30 @@
...
+    <Link href="#">links</Link> even!
+  </>
+);
+
+export const Playground: Story<CalloutProps> = (args) => <Callout {...args} />;
+Playground.args = {
+  icon: 'wrench',
+  children: content,
+};
+
```
ajkl2533, line 26: please disable the playground visual test with
```
Playgroung.parameters = { screenshot: { skip: true } };

File: src/components/Callout/Callout.tsx [hunk 2]
```diff
@@ -0,0 +1,54 @@
...
+  flex-shrink: 0;
+  background-color: ${getColor(ColorTypes.info700)};
+  border-radius: ${getRadii('circle')};
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  font-size: ${getFontSize('mdPlus')};
+`;
+
+const Container = styled.div`
```
ajkl2533, line 23: you can change `div` to `Padbox` and add the padding size to the Container component and save one DOM node.

src/components/Callout/Callout.tsx
```diff
@@ -0,0 +1,54 @@
...
+      <Inline gap="md">
+        <IconContainer>
+          {typeof icon === 'string' ? (
+            <Icon color={ColorTypes.neutral0} name={icon} />
+          ) : (
+            icon
+          )}
+        </IconContainer>
+        <Text
+          as="span"
```
ajkl2533, line 40: Text is `span` by default no need to add `as` property

src/components/Callout/Callout.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+const Container = styled.div`
+  background-color: ${getColor(ColorTypes.info50)};
+  border-radius: ${getRadii('default')};
+`;
+
+const Callout: React.FC<CalloutProps> = ({ children, icon }) => (
+  <Container>
+    <Padbox paddingSize={SpaceSizes.md}>
+      <Inline gap="md">
```
ajkl2533, line 31: please use `SpaceSizes.md`

src/components/Callout/Callout.tsx
```diff
@@ -0,0 +1,54 @@
...
+          style={{ alignSelf: 'center' }}
+        >
+          {children}
+        </Text>
+      </Inline>
+    </Padbox>
+  </Container>
+);
+
+Callout.propTypes = {};
```
ajkl2533, line 52: please add propTypes:
```
{
  icon: PropTypes.oneOfType([
    PropTypes.node,
    PropTypes.oneOf<SSCIcons>(Object.values(SSCIconNames)),
  ]).isRequired,
}

File: src/components/Callout/Callout.tsx [hunk 6]
```diff
@@ -0,0 +1,54 @@
...
+  justify-content: center;
+  font-size: ${getFontSize('mdPlus')};
+`;
+
+const Container = styled.div`
+  background-color: ${getColor(ColorTypes.info50)};
+  border-radius: ${getRadii('default')};
+`;
+
+const Callout: React.FC<CalloutProps> = ({ children, icon }) => (
```
ajkl2533, line 28: We should have some default icons so we don't need to define new icons with every usage. Please follow the discussion in Figma.

ajkl2533 (commented):
Looks good with some minor improvements. We just need to figure out the default icon before merging. Also, the public export is missing from `src/components/index.ts`

src/components/Callout/Callout.tsx
ajkl2533: @julesbarr Let's add this icon in the DS icons
https://github.com/FortAwesome/Font-Awesome/blob/5.x/js/regular.js#L230
https://fontawesome.com/v5/icons/lightbulb?s=regular

and  use it as the default icon

PR #503, opened 2022-06-06
feat: add input group

julesbarr:


src/components/InputGroup/InputGroup.tsx
```diff
@@ -0,0 +1,73 @@
...
+    display: block;
+  }
+`;
+
+const InputGroup: React.FC<InputGroupProps & InlineProps> = ({
+  children,
+  divider,
+  ...inlineProps
+}) => {
+  const ALLOWED_CHILDREN = [Select, Input, Icon, IconButton, InputGroup];
```
ajkl2533, line 56: I would remove IconButton (it is icon in the circle, it probably won't work with this) and add a normal `Button` instead. Let's also add `Password` and `MultiValueInput`

src/components/InputGroup/InputGroup.stories.tsx
```diff
@@ -0,0 +1,51 @@
...
+  <InputGroup stretch="end" divider>
+    <Select options={options} placeholder="Select a country" />
+    <InputGroup>
+      <Input key="input" placeholder="Search..." type="text" />
+      <Icon key="icon" name="search" />
+    </InputGroup>
+  </InputGroup>
+);
+
+NestedInputGroup.storyName = 'NestedInputGroup';
```
ajkl2533, line 51: Not needed

ajkl2533 (commented):
Looks good, please add default `index.ts` file in `InputGroup` folder and add `InputGroup` export in the `forms/index.ts`

src/components/forms/InputGroup/InputGroup.types.ts
```diff
@@ -0,0 +1,5 @@
+import { PropsWithChildren } from 'react';
+
+export type InputGroupProps = PropsWithChildren<{
+  divider?: boolean;
```
ajkl2533, line 4: I missed that  🤦 Could you please rename this to `hasDivider`? We are trying to keep bool props prefixed with is/has/should for better readabilty. Also, add it to prop types.

ajkl2533 (approved):
One prop rename but LGTM 👍

src/components/forms/InputGroup/InputGroup.types.ts
julesbarr: ✅  the only outstanding thing I didn't implement was support for MultiValue input, since this input can have more than one row.
I think we should add that one in a subsequent version since we'll need to adjust the design accordingly.

PR #508, opened 2022-06-08
fix(Button): add warning for href and disabled prop

julesbarr:


src/components/_internal/BaseButton/BaseButton.tsx
```diff
@@ -49,6 +49,14 @@ const BaseButton: React.FC<
     RouterLink = requireRouterLink();
   }
 
+  useEffect(() => {
```
ajkl2533, line 52: @julesbarr I don't think we need to use `useEffect` for this. let's replace that with 
```
if (process.env.NODE_ENV !== 'production') {
    console.warn('"isDisabled" prop in <Button> component will be ignored if the "href" prop is defined');
  }

ajkl2533 (approved):

PR #512, opened 2022-06-09
feat(Link): update colors and retire danger color option

Mnau-mnau:
implements https://zitenote.atlassian.net/browse/UXD-641


src/theme/typography.ts
```diff
@@ -77,21 +77,17 @@ const margin: Margin = {
...
+    decoration: 'underline',
   },
   secondary: {
-    color: themeColors.neutral[700],
-    hoverColor: themeColors.primary[400],
-    activeColor: themeColors.primary[200],
+    color: themeColors.neutral[1000],
+    hoverColor: themeColors.primary[800],
+    activeColor: themeColors.primary[1000],
+    focusBgColor: themeColors.primary[200],
```
ajkl2533, line 90: Those seems to be different from Figma. It should use neutral colors instead of primary

Mnau-mnau: changed

Mnau-mnau:
> 

Updated

PR #515, opened 2022-06-13
feat(Searchbar): set new icon positions for SearchBar

dclerici-ssc:
This PR changes the icon positions for SearchBar. 
Basically: 

- I’ve updated the search bar component design as follows
- the search icon is on the left
- when the field has input, a clear button appears on right while processing, and the search icon is replaced by a loading spinner


ajkl2533 (approved):
LGTM

PR #519, opened 2022-06-15
fix: let Select component recieve name property

mishalov:


ajkl2533:
Closing this as we already pass all the props into the Select component

PR #523, opened 2022-06-16
fix(Stepper): align label with bullet points

julesbarr:


ajkl2533 (approved):
Please add `display: block` to the Bullet component in https://github.com/securityscorecard/design-system/blob/fea516472975178b77e40d93bb808a1fd55585e1/src/components/Stepper/StepBullet.tsx#L25-L29

ajkl2533 (approved):
LGTM

PR #525, opened 2022-06-22
feat(Datatable): add visibility toggles to columns controls

ajkl2533:
This PR adds an option to show or hide columns through the columns control.

Breaking change:
- there was change in the properties:
    - `hasColumnOrdering` => `hasColumnsControls`
    - `defaultIsColumnOrderingOpen` => `defaultIsColumnsControlsOpen`
    - `defaultIsColumnOrderingApplied` =>  `defaultIsColumnsControlsApplied`
 
Other changes:
- the sortable list in columns control dropdown is now limited to 300px height with scrolling
- items in columns control can be dragged just by the handler, this was restricted so we can add switch to toggle visibility

Closes UXD-279


src/components/Datatable/Table/Table.tsx
```diff
@@ -200,6 +209,12 @@ TableProps<D> & { pageButtonsCount?: number }): React.ReactElement {
     collectFetchParams(newPageIndex, pageSize, sortBy);
   };
 
+  const resetHiddenColumns = () => {
+    DatatableStore.update((s) => {
+      s.hiddenColumns = [];
```
abudayssc, line 214: We could also reset the order here as it feels more like a "take me home" action

src/components/Datatable/components/ColumnsControls/ColumnsControls.tsx
```diff
@@ -55,6 +73,28 @@ const ColumnsControls: React.FC<ColumnsControlsProps> = ({
...
+            return (
+              <Inline
+                align="center"
+                data-id={id}
+                gap={SpaceSizes.sm}
+                justify="space-between"
+              >
+                <Text size={TextSizes.md}>{label}</Text>
+                <Switch
+                  checked={hiddenColumns.indexOf(id) === -1}
```
abudayssc, line 87: probably nicer to write `!hiddenColumns.includes(id)`

src/components/Datatable/components/ColumnsControls/hooks/useColumnVisibility.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    return () => {
+      unsubscribe();
+    };
+  }, []);
+  const setHiddenColumn = (id, value) => {
+    if (value === true) {
+      setLocalHiddenColumns((prev) => without([id], prev));
+    } else {
+      setLocalHiddenColumns((prev) => [...prev, id]);
```
abudayssc, line 36: Not sure what our current stance on ramda is but I'd either use a filter instead of `without` or use a proper functional expression to create the array. Just want to avoid using ramda as a toolkit library. But if we're okay with this, no arguments here.

src/components/SortableList/SortableList.tsx
```diff
@@ -16,15 +17,27 @@ import {
...
+import styled from 'styled-components';
 
 import { Stack } from '../layout';
 import { SpaceSizes } from '../../theme';
 import SortableItem from './SortableItem';
 import { SortableListProps } from './SortableList.types';
+import { pxToRem } from '../../utils';
+
+const SortableListRoot = styled.div<{
+  $maxHeight: SortableListProps['maxHeight'];
```
abudayssc, line 29: 🙏 this makes it feel so much nicer to use

abudayssc (approved):
Looks and works great @ajkl2533 💯. Some general comments, that are more about code than functionality. 
Also, I've been thinking about whether we want to let our users in the long run to specify these columns on their machine for future use, allowing them to have a custom view so that if they prefer to hide/reorder some columns they don't have to do it all the time.

src/components/Datatable/components/ColumnsControls/hooks/useColumnVisibility.ts
ajkl2533: Ramda is 🔥  🤣  I'm fine with using `filter` in imperative code here. I used `without` because it was shorter 😄

ajkl2533:
@abudayssc updated PR
About:
> specify these columns on their machine for future use
this was originally planned but then it was descoped now it's on the table again, but we are currently having discussion it it should be part of the component itself or it should be implemented by end-user

PR #526, opened 2022-06-23
fix(Button): change focus state behaviour

dclerici-ssc:
This PR fixes the `:focus` to `:focus-visible` for the `Button` component and for the buttons at `DatePicker/DatePickerRange.` 

The rest of the components don't need an update because the usage of the focus is correct. 

[Closes](https://zitenote.atlassian.net/browse/UXD-636)


Mnau-mnau (approved):

ajkl2533 (approved):
LGTM

PR #530, opened 2022-06-27
feat: remove shadow from react-select after Icon in InputGroup

mishalov:


ajkl2533 (approved):
Please generate missing visual test and update commit message to `fix(InputGroup): remove border when Select is not first element`

PR #536, opened 2022-07-01
feat(Button): add loading prop and extract renderButton function

julesbarr:


src/components/SemanticModal/SemanticModal.tsx
```diff
@@ -50,27 +70,11 @@ const SemanticModal = forwardRef<HTMLDivElement, SemanticModalProps>(
       onClose,
       variant = SemanticModalVariants.success,
       primaryButtonColor = ButtonEnums.ButtonColors.primary,
+      isPrimaryButtonLoading = false,
+      renderButton = defaultRenderButton,
```
ajkl2533, line 74: Hi @julesbarr, what is the reason for exposing `renderButton` property?

julesbarr: I might missunderstood the ticket requirements, but isn't this what is being asked in the ticket description?

ajkl2533: Ah ok, it was meant just to move it outside the component method so se do not recreate the function with every rerender. So it would be completely fine to just drop the property and it’s done. 👍

ajkl2533 (approved):

PR #540, opened 2022-07-12
feat: add overview page

julesbarr:


src/overview.stories.mdx
```diff
@@ -0,0 +1,433 @@
...
+import { useState } from 'react';
+import { ButtonColors } from './components/Button/Button.enums';
+import { noop, sign } from 'ramda-adjunct';
+import { BadgeSizes } from './components/Badge/Badge.enums';
+import { BrowserRouter } from 'react-router-dom';
+import { themes } from '@storybook/theming';
+
+<Meta title="Overview" />
+
+export const StickerContainer = styled.a`
```
ajkl2533, line 68: @julesbarr this being an `a` tag causes I'm being immediately redirected to some other page. If I change this to `div` it works fine.

src/overview.stories.mdx
```diff
@@ -0,0 +1,433 @@
...
+          onDragOver={function noRefCheck() {}}
+          onDragStart={function noRefCheck() {}}
+          onOrderChange={function noRefCheck() {}}
+        />
+      </StickerItem>
+      <StickerItem title="Spinner" id="components-spinner">
+        <Spinner dark />
+      </StickerItem>
+      <StickerItem title="Stepper" id="components-stepper">
+        <Stepper activeStep={0}>
```
ajkl2533, line 392: Stepper is not showing. It says it has no children

src/overview.stories.mdx
ajkl2533: I would probably add a button on the bottom of the card with the text "See documentation". It will be also better because users cannot interact with component in the cards if the whole card acts as a button

src/overview.stories.mdx
```diff
@@ -0,0 +1,483 @@
...
+          variant={HeadingVariants.secondary}
+          size={HeadingSizes.h5}
+        >
+          {title}
+        </Heading>
+        {children}
+      </Stack>
+    </StickerContainer>
+  );
+};
```
ajkl2533, line 177: So the issue seems to be that Storybook kicks up its own router only when the `a` tag is directly used and otherwise it causes issues. I suggest to copy this over your code:
```jsx
export const StickerItem = ({ title, children, id }) => {
  return (
    <StickerContainer>
      <Stack
        style={{ minHeight: 300 }}
        align="space-between"
        gap={SpaceSizes.md}
      >
        <Heading
          variant={HeadingVariants.secondary}
          size={HeadingSizes.h5}
        >
          {title}
        </Heading>
        <Inline justify="center">
          {children}
        </Inline>
        <Inline justify="flex-end">
          <a href={`?path=/docs/${id}--playground`}>See documentation →</a>
        </Inline>
      </Stack>
    </StickerContainer>
  );
};
```
I slightly changed the layout and added the `See documentation` button on the bottom

src/overview.stories.mdx
```diff
@@ -0,0 +1,483 @@
...
+
+export const StickerList = () => {
+  return (
+    <Grid cols={3} gap={SpaceSizes.lg}>
+      <StickerItem title="Accordion" id="components-accordion">
```
ajkl2533, line 182: We will need to update `id` properties to contain the whole story name. Good news is that except 4 components we can just add `--playground` at the end of the id (e.g. `components-accordion--playground`).

For Label, CloseButton, Icon, and Pagination you will need to check names in Storybook.

ajkl2533 (approved):
Please update commit message to `docs: add overview page` and merge it 👍

PR #548, opened 2022-07-19
feat(Card): add adornments

julesbarr:


src/components/Card/CardContent.tsx
```diff
@@ -1,11 +1,19 @@
 import React from 'react';
 
+import { SpaceSizes } from '../../theme';
+import { CardContainer } from './Card';
+
 const CardContent = React.forwardRef<
   HTMLDivElement,
   React.PropsWithChildren<Record<string, unknown>>
 >(({ children, ...props }, ref) => (
   <div ref={ref} {...props}>
```
ajkl2533, line 10: Let's move `ref` and `...props` to the `CardContainer` there is no need to add another node into the DOM

src/components/Card/CardHeader.tsx
```diff
@@ -57,49 +68,61 @@ const CardHeader = React.forwardRef<HTMLDivElement, CardHeaderProps>(
...
+  ),
 );
 
 CardHeader.propTypes = {
   title: PropTypes.string.isRequired,
-  adornment: PropTypes.node,
   subtitle: PropTypes.string,
   actions: PropTypes.arrayOf(ActionKindsPropType),
   actionsButtonLabel: PropTypes.string,
+  leftAdornment: PropTypes.oneOfType([PropTypes.node, PropTypes.string]),
```
ajkl2533, line 122: I think `PropTypes.node` should be enough IIRC it also cover the strings

src/components/Card/CardHeader.tsx
```diff
@@ -57,49 +68,61 @@ const CardHeader = React.forwardRef<HTMLDivElement, CardHeaderProps>(
...
               <ActionButton
                 aria-label={actionsButtonLabel}
                 title={actionsButtonLabel}
               >
-                <IconWrapper>
+                <IconAdornmentWrapper
+                  paddingSize={SpaceSizes.md}
+                  paddingType={PaddingTypes.squish}
+                >
                   <Icon name={SSCIconNames.ellipsisV} />
```
ajkl2533, line 106: I missed that also. But the button should be now using the horizontal ellipsis icon.

src/components/Card/Card.stories.tsx
```diff
@@ -63,17 +86,26 @@ function CardHeaderTemplateWithAdornment() {
 function CardContentTemplate() {
   return (
     <CardContent>
-      <Paragraph margin="none">
+      <Text size="md">
```
ajkl2533, line 89: The `Paragraph` component here was correct. `Text` is rendering span by default. It does not matter that much in this case but we should give correct examples to others.

src/components/Card/Card.stories.tsx
```diff
@@ -98,7 +130,7 @@ Playground.args = {
   children: (
     <>
       <CardMediaTemplate />
-      <Stack gap={SpaceSizes.md}>
+      <Stack>
```
ajkl2533, line 133: We can drop the `Stack` from here completely

src/components/Card/Card.types.ts
```diff
@@ -12,17 +13,14 @@ export interface CardActionsProps {
    * List of available actions for the Card footer (maximal number of actions is 2)
    */
   actions: ActionKinds<React.MouseEvent[]>[];
+  rightAdornment: React.ReactNode;
```
ajkl2533, line 16: Let's add comment here
```
/**
  * Adornment placed on the right side.
  */

File: src/components/Card/Card.types.ts [hunk 7]
```diff
@@ -35,6 +33,22 @@ export interface CardHeaderProps {
    * Actions menu aria label
    */
   actionsButtonLabel?: string;
+  /**
+   * Adornment placed on the left side.
+   */
+  leftAdornment?: React.ReactNode;
+  /**
+   * Adornment placed on the right side.
```
ajkl2533, line 41: Change to:
```
Callback to make action on click to help button in the Card title. If not provided no button will be rendered

File: src/components/Card/CardHeader.tsx [hunk 8]
```diff
@@ -1,31 +1,23 @@
...
+import { CardContainer } from './Card';
+import { PaddingTypes } from '../layout/Padbox/Padbox.enums';
 
-const ActionsWrapper = styled.div`
-  && {
-    margin-top: calc(${getSpace(SpaceSizes.xs)} * -1);
-    margin-right: calc(${getSpace(SpaceSizes.xs)} * -1);
-  }
-`;
+const ActionsWrapper = styled.div``;
```
ajkl2533, line 18: Let's remove this styled component and use just a simple `div` element in code

src/components/Card/CardHeader.tsx
```diff
@@ -1,31 +1,23 @@
...
 
-const IconWrapper = styled.div`
-  width: 2rem;
-  height: 2rem;
-  display: flex;
-  align-items: center;
-  justify-content: center;
+export const IconAdornmentWrapper = styled(Padbox)<{ width?: string }>`
```
ajkl2533, line 20:
```suggestion
export const IconAdornmentWrapper = styled(Padbox)`
```

src/components/Card/Card.types.ts
```diff
@@ -35,6 +33,22 @@ export interface CardHeaderProps {
...
+   */
+  onHelpClick?: () => void;
+  /**
+   * Title max number of lines
+   */
+  titleLinesCount?: number;
+  /**
+   * Subtitle max number of lines
+   */
+  subtitleLinesCount?: number;
```
ajkl2533, line 51: Let's use `maxTitleLinesCount` and `maxSubtitleLinesCount` it is more consistent wit similar cases we already have in DS

ajkl2533 (approved):
LGTM please resolve comments, update visual tests and change commit message to:
`feat(Card): align Card styling with new design`

PR #561, opened 2022-08-08
fix(Card): make Card adjustments

ajkl2533:
1. fixed alignment of adornment, title, and buttons in the CardHeader
2. help button can show a tooltip if the element is passed into `helpTooltip` property
3. the last element inside `Card` is aligned to the bottom edge of the card
4. horizontal card variant was removed

H4 and H5 line heights were decoupled from paragraph line height and can be now used with the correct line height with Inter font.


Mnau-mnau (approved):
looks good

PR #581, opened 2022-08-24
Update icons to v6

ajkl2533:


MateCziner (commented):
reviewed the cases github could render, all those are fine

Mnau-mnau (approved):

PR #584, opened 2022-08-30
feat(SegmentedToggle): change font style to regular

Mnau-mnau:
implements https://zitenote.atlassian.net/browse/UXD-842

font size is already set to md


ajkl2533 (changes requested):
I'm not sure why this PR exist. The segmented toggle already uses regulat weight (see L107) and there is no need to move the default font-weight to each variant

PR #592, opened 2022-09-13
feat(BatchActions): make the caret down same color as text

Mnau-mnau:
implements https://zitenote.atlassian.net/browse/UXD-788


src/components/Datatable/BatchModule/BatchActions/BatchActions.tsx
```diff
@@ -80,7 +80,7 @@ const BatchActions: React.FC<BatchActionsProps> = ({ actions }) => {
                 <Inline align="center" gap={SpaceSizes.xs}>
                   <span>{action.label}</span>
                   <Icon
-                    color={ColorTypes.neutral700}
+                    color={ColorTypes.primary600}
```
ajkl2533, line 83: We should drop the `color` prop entirely. Icon component inherits parent color so it should be inheriting from the button

Mnau-mnau: removed

ajkl2533 (approved):

PR #598, opened 2022-09-30
refactor: add displayNames to minified components

ajkl2533:
Closes UXD-887

### DIFF
--- a/src/components/_internal/BaseTable/Head/Head.tsx
+++ b/src/components/_internal/BaseTable/Head/Head.tsx
@@ -83,4 +83,5 @@ function Head<D extends Record<string, unknown>>({
   );
 }
 
+Head.displayName = 'Head';
 export default Head;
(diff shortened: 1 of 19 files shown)


src/components/_internal/BaseTable/Head/Head.tsx
```diff
@@ -83,4 +83,5 @@ function Head<D extends Record<string, unknown>>({
   );
 }
 
+Head.displayName = Head;
```
julesbarr, line 86: there are  missing quote marks in this line

julesbarr (changes requested):

src/components/_internal/BaseTable/Head/Head.tsx
ajkl2533: 🤦 fixed

julesbarr (approved):

PR #606, opened 2022-10-06
feat: add className to components

Mnau-mnau:
implements https://zitenote.atlassian.net/browse/UXD-928


src/components/layout/Center/Center.tsx
```diff
@@ -65,12 +66,14 @@ Center.propTypes = {
...
   areChildrenCentered: PropTypes.bool,
   isTextCentered: PropTypes.bool,
+  className: PropTypes.string,
 };
 
 Center.defaultProps = {
   gutters: SpaceSizes.none,
   areChildrenCentered: false,
   isTextCentered: false,
+  className: CLX_LAYOUT,
```
ajkl2533, line 76: We cannot do it this way because default props are overridden from outside, you need to do it like: 
```ts
const Center = styled.div
  .withConfig({
    shouldForwardProp: (prop) =>
      !includes(prop, [
        'maxWidth',
        'gutters',
        'areChildrenCentered',
        'isTextCentered',
      ]),
  })
  .attrs((props) => ({
    ...props,
    className: classNames(CLX_LAYOUT, props.className),
  }))<CenterProps>

ajkl2533:
@Mnau-mnau couple things to fix:
- `Col`, `Container`, `Tooltip` - missing className
- `Spinner` - have className set in the defaultProps
- `Strong`, `Code` - inherits className from `Text` but also adds its own see there are two same classNames in the code
- `SemanticModal`, `WizardModal` - same issue as above, just inheriting from `Modal`
and there is a conflict in `Switch`

Mnau-mnau:
> @Mnau-mnau couple things to fix:
> 
> * `Col`, `Container`, `Tooltip` - missing className
> * `Spinner` - have className set in the defaultProps
> * `Strong`, `Code` - inherits className from `Text` but also adds its own see there are two same classNames in the code
> * `SemanticModal`, `WizardModal` - same issue as above, just inheriting from `Modal`
>   and there is a conflict in `Switch`

Tooltip uses Dropdown and Dropdown has the className on it

ajkl2533 (approved):
LGTM

PR #610, opened 2022-10-12
docs(Datatable): add icon to batch actions

julesbarr:


ajkl2533:
@julesbarr Looks OK to me, but I would move the example into one of the playground stories we have for Datatable. I don’t expect that people will be able to find it hidden deep within the `internalComponent` folder

ajkl2533 (approved):
LGTM

PR #611, opened 2022-10-12
fix(Heading): decouple size from tag

julesbarr:


src/components/typographyLegacy/Heading/Heading.types.tsx
```diff
@@ -6,5 +6,6 @@ export type Variants = typeof HeadingVariants[keyof typeof HeadingVariants];
 
 export interface HeadingProps extends SpacingProps {
   size?: Sizes;
+  as?: string;
```
ajkl2533, line 9: Seems that this type annotation update is not even needed. I would remove it and change the commit message to:
`docs(Heading): add custom element example` we wouldn't need to release a new version of DS. 

I'm not sure why I was convinced that this is not working properly 🤦

PR #612, opened 2022-10-12
fix(Card): add interactive properties

julesbarr:


ajkl2533 (approved):
LGTM, please change the commit message to `feat(Card): add interactive properties`

src/components/Card/Card.stories.tsx
```diff
@@ -372,3 +373,37 @@ export const LastItemAlignment = () => (
...
   </Grid>
 );
+
+const CustomLink = styled.a`
+  border: 2px solid purple !important;
+`;
+
+export const InteractiveButtonCard = () => (
+  <Stack gap="lg">
+    <Heading size="h2">Interactive button card</Heading>
```
ajkl2533, line 383: Is there any specific reason to have the headings inside stories? If not I would remove those, story name should be enough.

PR #613, opened 2022-10-17
feat(Grid): throw Error on cols=1

Mnau-mnau:
implements https://zitenote.atlassian.net/browse/UXD-926

### DIFF
--- a/src/components/layout/Grid/Grid.tsx
+++ b/src/components/layout/Grid/Grid.tsx
@@ -75,13 +75,26 @@ const Grid: React.FC<GridProps> = ({
   wrapperEl,
   wrapperOverflow = 'hidden',
   ...props
-}) => (
-  <GridWrapper $overflow={wrapperOverflow} as={wrapperEl}>
-    <GridParent $align={align} $cols={cols} $gap={gap} as={parentEl} {...props}>
-      {children}
-    </GridParent>
-  </GridWrapper>
-);
+}) => {
+  if (cols === 1) {
+    throw new Error(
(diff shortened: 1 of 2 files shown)


src/components/layout/Grid/Grid.tsx
```diff
@@ -75,13 +75,26 @@ const Grid: React.FC<GridProps> = ({
...
   ...props
-}) => (
-  <GridWrapper $overflow={wrapperOverflow} as={wrapperEl}>
-    <GridParent $align={align} $cols={cols} $gap={gap} as={parentEl} {...props}>
-      {children}
-    </GridParent>
-  </GridWrapper>
-);
+}) => {
+  if (cols === 1) {
```
ajkl2533, line 79: I'd probably use `less than 2` instead of `equals 1`

src/components/layout/Grid/Grid.tsx
```diff
@@ -75,13 +75,26 @@ const Grid: React.FC<GridProps> = ({
...
+    throw new Error(
+      '[design-system/Grid] Minimal number of columns is 2. Use Stack instead of ”Grid[cols=1]',
```
ajkl2533, line 81: Seems that the double quote does not end anywhere 🤔 😄

src/components/layout/Grid/Grid.tsx
Mnau-mnau: cols < 2 breaks Storybook, so keeping it as it is, or until some creative person starts using cols={0}

PR #616, opened 2022-10-24
fix(BatchActions): add tooltip option

julesbarr:


src/components/Datatable/BatchModule/BatchActions/BatchActions.tsx
```diff
@@ -65,7 +67,7 @@ const BatchActions: React.FC<BatchActionsProps> = ({ actions }) => {
             (action as ActionWithSubactions<BatchActionArgs>).subActions,
           );
 
-          return (
+          element = (
```
ajkl2533, line 70: We have this conditional tooltip wrapper which could save us this reassigning; check the https://github.com/securityscorecard/design-system/blob/alpha/src/components/_internal/BaseTable/components/TooltipWrapper.tsx

julesbarr: I noticed the tooltip component has the same functionality https://github.com/securityscorecard/design-system/blob/alpha/src/components/Tooltip/Tooltip.tsx#L20
The reassigning is because the component render different elements if it has subactions defined

ajkl2533: Right 👍

ajkl2533 (approved):

ajkl2533:
@julesbarr could you please change the commit message from `fix` to `feat`

PR #618, opened 2022-10-26
fix(Datatable): set rows action pane width to auto

julesbarr:


src/components/_internal/BaseDropdownMenu/DropdownMenu.tsx
```diff
@@ -30,6 +30,7 @@ export const DropdownLink = styled(Padbox).withConfig({
   shouldForwardProp: (prop) => !FORWARD_PROP_EXCEPTIONS.includes(String(prop)),
 })<DropdownLinkProps>`
   width: 100%;
+  min-width: 140px;
```
ajkl2533, line 33: Should be in rem

src/components/_internal/BaseTable/columns/actionsColumn.tsx
```diff
@@ -50,15 +50,19 @@ export const actionsColumn = {
...
       ...action,
       onClick: () => action.onClick(row.id, row.original),
     }))(rowActions);
 
     return (
       <Inline justify="center" stretch="start">
-        <DropdownMenu actions={actions} paneWidth={140} placement="bottom-end">
+        <DropdownMenu
+          actions={actions}
+          paneWidth={rowActionsPaneWidth}
```
ajkl2533, line 63: This seems to be overly complicated. I was thinking about setting the `paneWidth="auto"` and letting the content decide on how wide the menu should be. It should also limit the number of errors in tests :)

julesbarr: that's the default value, 'auto' and it's auto-adjusting based on the content

src/components/_internal/BaseDropdownMenu/DropdownMenu.tsx
julesbarr: 👍

src/components/_internal/BaseTable/columns/actionsColumn.tsx
ajkl2533: Yeah I know I don't want to pollute Datatable component with another property if we can do the sizing automatically. I think that DT is already too complicated

ajkl2533:
@julesbarr could you please change the commit message from `fix` to `feat`

ajkl2533 (approved):

PR #622, opened 2022-10-28
feat(Wizard): add Wizard

julesbarr:


OldrichKruchna (commented):
Great effort @julesbarr, so far looks good to me, keep going!

src/components/Wizard/Wizard.context.tsx
```diff
@@ -0,0 +1,59 @@
...
+  SetStateAction,
+  useEffect,
+  useMemo,
+  useState,
+} from 'react';
+
+import { WizardStep } from './Wizard.types';
+
+export interface WizardState {
+  initialStep?: string;
```
marekpukaj, line 12: We need to set not just an initial step but also initial history. Our temporary solution for this looks like this https://github.com/securityscorecard/ssc-ui-core/pull/10797/files#diff-82c927276cacb46cf568f776be331b315e999f183175a6bb8ee87b4a9c0532e7L13-R17

ajkl2533: What is the initial history?

marekpukaj: Initial history is basically the set of completed steps. Looking at this wizard, it seems like it doesn't keep a track of completed steps which is something that we need. Setting just an initial state doesn't mean that all prev steps have been completed

ajkl2533: Well but that is not something supported by Stepper anyway. How would you represent that? Stepper expects that all previous steps are always completed. It also doesn't make much sense, why would you bring a user to a step that expects some previous steps to be completed but actually aren't? Maybe I don't understand the use case, but also this is the first time I'm hearing about this.

marekpukaj: The specific use case for this is for example invite flow wizard. You can start from the first step which is add the company and then continue with the rest of the steps. Or you can start in the middle of the flow in invite company step (if you click on invite company button) and at the end of the flow you can add another company which brings you to the add company step.

ajkl2533: Well but that means that you have the data for the first step and thus the step can be considered complete. I don't get why you need a history array. What that brings you?

marekpukaj: We can check if the state was visited and based on that modify the logic. But I agree we can handle this outside of wizard.

marekpukaj (approved):
LGTM, thank you for implementing this

PR #625, opened 2022-11-01
fix(UserAvatar): add isInverted prop

julesbarr:


src/components/UserAvatar/UserAvatar.types.ts
```diff
@@ -13,4 +13,8 @@ export interface UserAvatarProps {
    */
   size?: Sizes;
   className?: string;
+  /**
+   * Used to inverse the colors of the avatar
+   */
+  isInversed?: boolean;
```
ajkl2533, line 19: @julesbarr could you please change the prop name to `isInverted`?

ajkl2533 (approved):
LGTM, just please change `isInversed` in commit message to `isInverted` so we have correct prop name in Changelog

PR #628, opened 2022-11-07
fix(HexGrade): prop rename isInversed to isInverted

julesbarr:


Mnau-mnau (approved):

ajkl2533 (approved):
LGTM, but it's a breaking change. We need to be careful when upgrading the DS in platform

PR #632, opened 2022-11-11
fix(collapsible): updated header content

alehkonan:
Motivation of this PR is the possibility to use custom components in the header as on the screen below
Screenshot from 2022-11-11 16-58-28

- As far as title is a React.Node I changed a container from "span" to "div" as there can be other tags inside
- Added HeaderContent component with "flex: 1" to use all width of the header
- set subject prop as optional to avoid unnecessary margin if subject is empty string


ajkl2533 (commented):
LGTM, please do two more things:
1. update the commit message to `feat(Collapsible): allow custom elements in header`
2. fix failing visual tests

PR #653, opened 2022-12-06
docs(Breadcrumbs): improve the documentation

Mnau-mnau:
I found the lack of Breadcrumb details while putting together onboarding documentation, so here it is 🍏


julesbarr (approved):

src/components/Breadcrumbs/Breadcrumbs.types.ts
```diff
@@ -6,6 +6,8 @@ export interface BreadcrumbItemProps extends Omit<LinkProps, 'color'> {
    */
   children: React.ReactNode;
   isSelected?: boolean;
+  to?: string;
```
ajkl2533, line 9: Also, here is the wrong type:
```
import type { To } from 'history';

to?: To;

ajkl2533 (approved):
LGTM

PR #656, opened 2022-12-08
fix(Banner): move props to wrapper component

Mnau-mnau:


src/components/Banner/Banner.tsx
```diff
@@ -81,7 +81,9 @@ const Banner: React.FC<BannerProps> = ({
     <StyledPadbox
       $variant={variant}
       className={cls(CLX_COMPONENT, className)}
+      data-test-id={props?.['data-test-id']}
```
ajkl2533, line 84: that's unnecessary

ajkl2533 (approved):

PR #665, opened 2022-12-19
docs(Checkbox): change role and add descriptions to Storybook

Mnau-mnau:
implements https://zitenote.atlassian.net/browse/UXD-1010


julesbarr (approved):

ajkl2533:
This shouldn't be a "docs" commit as you are actually changing the implementation

Mnau-mnau:
> This shouldn't be a "docs" commit as you are actually changing the implementation

split to 2 commits, changed messages accordingly

ajkl2533 (approved):

PR #669, opened 2022-12-21
feat(Button): add aria-label to icon only buttons

Mnau-mnau:
implements https://zitenote.atlassian.net/browse/UXD-1014

[UXD-1014]: https://zitenote.atlassian.net/browse/UXD-1014?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/_internal/BaseButton/BaseButton.tsx
+++ b/src/components/_internal/BaseButton/BaseButton.tsx
@@ -52,7 +52,6 @@ const BaseButton: React.FC<
   isExpanded = false,
   loadingText = 'Loading',
   className,
-  ariaLabel,
   ...props
 }) => {
   let RouterLink = null;
@@ -94,7 +93,7 @@ const BaseButton: React.FC<
     </>
   ) : isNotUndefined(iconName) ? (
     <>
-      <BaseStyledIcon aria-label={ariaLabel} name={iconName} type={iconType} />
+      <BaseStyledIcon name={iconName} type={iconType} />
       {isNotUndefined(children) && <span>{children}</span>}
     </>
   ) : (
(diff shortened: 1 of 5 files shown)


src/components/_internal/BaseButton/BaseButton.tsx
```diff
@@ -52,6 +52,7 @@ const BaseButton: React.FC<
   isExpanded = false,
   loadingText = 'Loading',
   className,
+  ariaLabel,
```
ajkl2533, line 55: I think it would be better not to introduce a custom prop name and rather destructure `aria-label` prop from the `...props` and use it where we need. That way people wouldn't need to remember that button has a special name for this.

src/components/_internal/BaseButton/BaseButton.tsx
```diff
@@ -108,6 +108,7 @@ const BaseButton: React.FC<
       $isLoading={isLoading}
       $margin={margin}
       $variant={variant}
+      aria-label={props['aria-label']}
```
ajkl2533, line 111: So this actually doesn't solve the issue described in UXD-1014. The thing is that we need to provide our own `aria-label` for cases when the button has `isLoading === true && hasOnlyIcon === true`. In other cases, it is up to the consumer to provide `aria-label`. We should also move the line under the `{...props}` so this won't get overridden by the user-passed `aria-label` property.

```suggestion
      aria-label={isLoading && hasOnlyIcon ? 'Loading' : props?.['aria-label']}
```

ajkl2533 (approved):

PR #672, opened 2022-12-22
feat(CardMedia): make alt attribute required

Mnau-mnau:
implements https://zitenote.atlassian.net/browse/UXD-1016

BREAKING CHANGE: alt attribute needs to be always defined

[UXD-1016]: https://zitenote.atlassian.net/browse/UXD-1016?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Card/CardMedia.tsx
+++ b/src/components/Card/CardMedia.tsx
@@ -64,9 +64,9 @@ function CardMedia<El extends HTMLElement = HTMLDivElement>({
 }
 
 CardMedia.propTypes = {
+  alt: PropTypes.string.isRequired,
   mediaSrc: PropTypes.string,
   style: PropTypes.shape({}),
-  alt: PropTypes.string,
   as: PropTypes.elementType,
 };
 CardMedia.displayName = 'CardMedia';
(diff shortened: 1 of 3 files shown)


julesbarr (approved):

src/components/Card/CardMedia.tsx
```diff
@@ -50,7 +50,7 @@ function CardMedia<El extends HTMLElement = HTMLDivElement>({
     <CardMediaWrapper
       $isImageComponent={isImageComponent}
       $isMediaComponent={isMediaComponent}
-      alt={as === 'img' && hasMediaSrc ? alt : undefined}
+      alt={alt}
```
ajkl2533, line 53: I think that the condition should stay there as the `alt` attribute doesn't exist on tags other than `img`.  This way the `alt` attribute will be always there when the property is defined it's defined

ajkl2533:
@Mnau-mnau There is an exclamation mark in the commit message for some reason, could you please remove it?

ajkl2533 (approved):

PR #673, opened 2022-12-28
feat(Table): add focus to scrollable area

Mnau-mnau:
So there is focus added, but it's not perfect - at the datatable the outline is hidden under the sticky column and it looks weird. 
Removing the background shows the scrolled stuff underneath and it looks weird.
The only thing that so far helps is to add margins on focus so the outline would be exposed, but it also shifts its contents.

Ideas appreciated.

Screenshot 2022-12-28 at 11 28 46
Screenshot 2022-12-28 at 11 28 26


ajkl2533:
@Mnau-mnau the problem is that you are adding `tabIndex` on the wrong element it should be added to `BaseTableContainer` component, not to the `tbody` tag. That way the focus ring will be correctly around whole scrollable area not just around the data rows

ajkl2533 (approved):

PR #677, opened 2023-01-03
fix(Wizard): fix wizard state management

julesbarr:


OldrichKruchna (approved):
This is great! Thank you @julesbarr 🙇

PR #691, opened 2023-01-24
feat(UserAvatar): render non-interactive element

ajkl2533:


src/components/UserAvatar/UserAvatar.tsx
```diff
@@ -21,13 +29,22 @@ const mediumAvatar = css`
...
   color: ${getColor('neutral.1000')};
   background-color: ${getColor('neutral.0')};
   &:hover {
     background-color: ${getColor('primary.50')};
     color: ${getColor('neutral.900')};
   }
+  &:focus {
+    background-color: ${getColor('primary.800')};
+    color: ${getColor('neutral.0')};
+  }
```
Mnau-mnau, line 47: this seems off - inverted style has mainly light background and dark font except for this one - it is not defined in Figma, but I think it should be closer to the pressed state https://www.figma.com/file/vnzYfrwIISRYkr2iYE5TcD/DS-Atoms?node-id=17126%3A23042&t=e4WJ1u7Fq7zGqQlI-0

ajkl2533: Fair enough I changed it to white background with the focus ring. So it's similar to the outline button

PR #697, opened 2023-01-30
fix(Datatable): fix persisted state

julesbarr:


ajkl2533 (approved):
Tested in platform, looks good

PR #698, opened 2023-01-31
POC fix(Dropdown): implement focus trap

julesbarr:


ajkl2533:
Two things here:
1) I would extract the focus trap into separate hook under `src/hooks`
2) Your implementation causes the first interactive element in the dropdown automatically gain focus I don't think that is correct behavior. When the user opens a dropdown nothing should be prefocused we should wait for the `Tab` keypress. This also makes the visual tests fail

src/components/Dropdown/ControlledDropdown.tsx
```diff
@@ -25,6 +26,7 @@ const ControlledDropdown = forwardRef<
       maxPaneWidth = 270,
       isPaneElevated = false,
       hasPaneArrow = false,
+      focusTrap = false,
```
ajkl2533, line 29: Let's change this to `true` and then add `focusTrap=false` to the Tooltip component. Then we can merge this

PR #704, opened 2023-02-14
feat(Accordion): add external opening

Mnau-mnau:


julesbarr (approved):

src/components/Accordion/Accordion.tsx
```diff
@@ -18,50 +19,74 @@ const pickOpen: (items: AccordionItem[]) => AccordionItemId[] = pipe(
...
-          isOpen={includes(id, openIds)}
-          title={title}
-        >
-          {content}
-        </AccordionCollapsible>
-      ))}
-    </Stack>
-  );
-};
+    useEffect(
```
ajkl2533, line 50: `useDeepCompareEffect` might be better in this case since the `openItems` is an array and `useEffect` is checking reference equality so even if items are the same in the current and previous state the effect will be triggered in every rerender

ajkl2533 (approved):
Two small suggestions

PR #707, opened 2023-03-02
feat(Banner): add wide layout option

Mnau-mnau:
https://zitenote.atlassian.net/browse/UXD-1090


src/components/Banner/Banner.tsx
```diff
@@ -52,15 +53,45 @@ const ContentWrapper = styled(Padbox)`
 
 /* stylelint-disable */
 const Text = styled(BaseText)<{ $variant?: BannerProps['variant'] }>`
-  max-width: 125ch;
+  max-width: 150ch;
```
ajkl2533, line 56: Be careful, `ch` unit can be tricky and not always corresponds with the real number of characters for given font. E.g. for Inter it should be `115ch` to get 150 characters per line. So let's change it to `115`
https://meyerweb.com/eric/thoughts/2018/06/28/what-is-the-css-ch-unit/

src/components/Banner/Banner.tsx
```diff
@@ -95,35 +136,19 @@ const Banner: React.FC<BannerProps> = ({
...
-                      onClick={action.onClick}
-                    >
-                      {action.label}
-                    </StyledButton>
-                  ))}
-                </Inline>
-              )}
-            </Stack>
+            {isInline ? (
+              <Inline align="flex-start" gap={SpaceSizes.md}>
```
ajkl2533, line 140: please add `justify: space-between` right now actions sticks to the content and it looks strange

src/components/Banner/Banner.tsx
```diff
@@ -52,15 +53,45 @@ const ContentWrapper = styled(Padbox)`
 
 /* stylelint-disable */
 const Text = styled(BaseText)<{ $variant?: BannerProps['variant'] }>`
-  max-width: 125ch;
+  max-width: 150ch;
   color: ${getColor('neutral.1000')};
   display: -webkit-box;
   overflow-y: auto;
```
ajkl2533, line 59: I think you can also remove those two lines and then remove the stylelint comment

ajkl2533 (approved):

PR #721, opened 2023-03-21
fix: remove cloneElement

julesbarr:


src/components/Breadcrumbs/BreadcrumbItem.tsx
```diff
@@ -31,9 +31,10 @@ const BreadcrumbItem: React.FC<BreadcrumbItemProps> = ({
 }) => {
   const hasIcon = isNotUndefined(iconName);
   const label = hasIcon ? { 'aria-label': children } : { children };
+  const isText = isNilOrEmpty(to) && isNilOrEmpty(href);
   return (
     <ListItem>
-      {isSelected ? (
+      {isSelected || isText ? (
```
ajkl2533, line 37: I guess we can remove the `isSelected` property since that was filled in the `cloneElement` and the implementation was exactly the same as the `isText`.

src/components/Tabs/Tabs.tsx
```diff
@@ -19,40 +30,38 @@ const Tabs: FC<TabsProps> = ({
...
-    $variant={variant}
-    className={CLX_COMPONENT}
-    paddingSize={
-      variant === TabVariants.segmented ? SpaceSizes.xs : SpaceSizes.none
-    }
+  <Provider
+    value={{
+      selectedPatternMatcher,
+      selectedValue,
+      onSelectTab,
```
ajkl2533, line 37:
```suggestion
      selectTab: onSelectTab,
```
I think we can change the `onSelectTab` property name to `selectTab` because we are not using it in the `Tab` component as a handler but a method to select tab so the `on` prefix is unnecessary

ajkl2533 (approved):
LGTM

PR #722, opened 2023-03-21
fix: change default font size to md

julesbarr:


ajkl2533 (approved):
LGTM. The visual tests are affected by recompilation of the testing docker image and are of to update

PR #724, opened 2023-03-22
feat(icons): add link icon

abudayssc:


src/theme/icons/link.ts
```diff
@@ -0,0 +1,17 @@
+import {
+  IconDefinition,
+  IconName,
+  IconPrefix,
+} from '@fortawesome/fontawesome-svg-core';
+
+export const width = 20;
```
Mnau-mnau, line 7: I don't know the icon source but the dimensions come from the SVG file sizing, most of them are around 512px, just make sure the icon displays correctly

Mnau-mnau (changes requested):
some suggestions

PR #730, opened 2023-03-28
fix(Filters): remove dupe input

julesbarr:


src/components/Filters/components/Number/Number.tsx
```diff
@@ -1,20 +1,10 @@
...
 import { NumberPropTypes, NumberProps } from './Number.types';
-
-export const StyledNumber = styled(StyledInput)`
-  &::-webkit-inner-spin-button,
-  &::-webkit-outer-spin-button {
-    appearance: none;
-    margin: 0;
-  }
-  appearance: textfield;
-`;
```
ajkl2533, line 17: @julesbarr can we migrate this to the Input component and maybe add a prop to toggle this behaviour? My concern is that some tables use this field and don't validate the negative numbers you can't write in but can set with those buttons.

something like:
```scss
[type=number][data-has-spin=false] { 
  &::-webkit-inner-spin-button,
  &::-webkit-outer-spin-button {
    appearance: none;
    margin: 0;
  }
  appearance: textfield;
}

julesbarr: I just restored that css in the input

ajkl2533 (approved):

PR #736, opened 2023-03-30
fix(Datatable): modify row actions rendering

julesbarr:


ajkl2533:
There is no non-programmatic way how to dynamically determine the width of the actions column based on the width of the button label. Due to this limitation, we decided to cancel this ticket

PR #758, opened 2023-04-21
feat(ProgressBar): add progress bar

julesbarr:


src/components/ProgressBar/index.tsx
```diff
@@ -0,0 +1 @@
+export { default as ProgressBar } from './ProgressBar';
```
ajkl2533, line 1: Please add `export * from './ProgressBar.types';`

ajkl2533 (approved):

PR #761, opened 2023-05-04
feat(Datatable): allow disabling batch action buttons

mrthebunny:


src/components/Datatable/BatchModule/BatchActions/BatchActions.stories.tsx
```diff
@@ -28,11 +28,14 @@ Menu.argTypes = {
         summary: 'Action[]',
         detail: `
 | {
+    isDisabled?: boolean;
     label: string;
     name: string;
+    disabled?: boolean;
```
ajkl2533, line 34: I guess this shouldn't be here.

src/components/Datatable/BatchModule/BatchActions/BatchActions.stories.tsx
```diff
@@ -49,13 +53,23 @@ Menu.argTypes = {
...
 | {
+    isDisabled?: boolean;
     label: string;
     name: string;
     subActions: ActionKinds[];
     onClick?: () => void;
     href?: never;
     to?: never;
   }
+  | {
```
ajkl2533, line 64: This is the same as the one above

src/types/action.types.ts
```diff
@@ -8,6 +8,7 @@ export type ActionBase<
   OnClickArgs extends Array<unknown>,
   OnClickReturnType = void,
 > = {
+  isDisabled?: boolean;
```
ajkl2533, line 11: You shouldn't be extending the generic ActionBase because suddenly all component using this type will be allowed to have `isDisabled` property even if the component itself is not using it. ActionKinds allowes to pass additional props in the third generic argument ActionWithSubactions does not yet so that will needs to be adjusted. But technically you should change the type in https://github.com/securityscorecard/design-system/blob/main/src/components/Datatable/BatchModule/BatchActions/BatchActions.types.ts

mrthebunny:
Closing this PR because UX guidelines prohibit disabling batch action buttons 
image

source: https://zitenote.atlassian.net/wiki/spaces/DSA/pages/3367305234/Guide%3A+Inactive+Buttons+and+Hidden+Workspaces?atlOrigin=eyJpIjoiMGM5NTFhYmNiZjY5NDFmYzg2N2IyZDliNjg2Mjg1NmUiLCJwIjoiY29uZmx1ZW5jZS1jaGF0cy1pbnQifQ

PR #762, opened 2023-05-09
feat(Pill): add new color palette

Mnau-mnau:
implements https://zitenote.atlassian.net/browse/UXD-1154


src/components/Pill/PillColors.ts
```diff
@@ -0,0 +1,28 @@
+export const PillColors = {
```
ajkl2533, line 1: I think this should go to enums, this file is not exported so nobody can use it outside DS.

src/components/Pill/PillColors.ts
```diff
@@ -0,0 +1,28 @@
...
+    turquoise: '#B8DFFA',
+    teal: '#A6CFCF',
+    green: '#B1DCBE',
+    yellow: '#FDF2B8',
+    orange: '#FFD0A6',
+    red: '#FAA6A6',
+    pink: '#EEC8F6',
+    purple: '#C9BCFB',
+    brown: '#D6B9A6',
+  },
```
ajkl2533, line 27: This looks like lot of repetition, can we simplify the format to:
```js
const PillColors = {
  gray: {default: '#E8E8E8', hover: '#D8D8D8' },
  blue: { default: ..., hover: ... },
  turquoise: ...,
```
or maybe even using tuple would be enough like
```js
const PillColors = {
  gray: ['#E8E8E8', '#D8D8D8'],
  blue: [default, hover],
  turquoise: ...,
```

src/components/Pill/PillWrapper.tsx
```diff
@@ -3,57 +3,45 @@ import styled, { css } from 'styled-components';
...
     css`
-      cursor: pointer;
-
+      background-color: ${PillColors.pill[$color]};
       &:hover,
       &:focus {
-        background-color: ${getColor('primary.50')};
+        background-color: ${PillColors.pillHover[$color]};
         outline: none;
       }
```
ajkl2533, line 24: This should go a bit lower into `$isClickable` condition. Right now even pills that are not clickable have hover state

src/components/forms/Select/Select.styles.tsx
```diff
@@ -297,9 +297,8 @@ export const MultiValueContainer: ComponentType<Record<string, unknown>> = ({
   ...props
 }) => (
   <PillWrapper
+    color={PillColors.gray}
```
ajkl2533, line 300: Why this? Gray is default value so this shouldn't be needed

src/components/Pill/PillColors.ts
Mnau-mnau: Isn't that exactly what we want? We don't want the color range in the theme so it wouldn't be spread all around outside the pills - we just want to offer enums to our consumers.

ajkl2533: We don't want this to be in theme, because we don't want to encourage people to use these colours for styling. But they need it to be able to build a tag color selector so it needs to be exported anyway

src/components/Pill/Pill.stories.tsx
```diff
@@ -67,11 +52,17 @@ Removable.args = {
...
+        color={color}
+        label={capitalize(color)}
+        onClick={action('OnRemove')}
+      />
+    ))}
+  </Inline>
+);
 
 export const ClickableAndDeletable = PillTemplate.bind({});
 ClickableAndDeletable.args = {
```
ajkl2533, line 68: You removed `args` definition for `Clickable` story which this story uses. Due to this the pill in this story is no longer clickable.

You can revert the changes for the `Clickable` story and just reuse `args` property for all instances like:
```jsx
export const Clickable: Story<PillProps> = (args) => (
  <Inline gap="md">
    {Object.values(PillColorsEnums).map((color) => (
      <Pill
        key={color}
        {...args}
        color={color}
      />
    ))}
  </Inline>
);

Clickable.args = {
  ...Playground.args,
  onClick: action('OnRemove'),
};

ajkl2533 (approved):

PR #770, opened 2023-05-24
feat: add score delta component

julesbarr:


src/components/ScoreDelta/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+  ({ delta, decimalsCount = 0 }, ref) => {
+    const icon = delta === 0 ? SSCIconNames.minus : SSCIconNames.arrowUp;
+    const type = getType(delta);
+    const coeficient = Math.floor(Math.abs(delta));
+    const decimals = getDecimals(delta, decimalsCount);
+    const text = useMemo(() => {
+      return [coeficient, decimalsCount > 0 ? decimals : undefined]
+        .filter((item) => item !== undefined)
+        .join('.');
+    }, [coeficient, decimals, decimalsCount]);
```
ajkl2533, line 73: This seems unnecessarily complex. You could just use:
`const label = Math.abs(Number.parseFloat(delta).toFixed(decimalsCount))`

ajkl2533:
Let's change the commit message to `feat(ScoreDelta): add component`

src/components/ScoreDelta/index.tsx
```diff
@@ -0,0 +1,73 @@
+import React from 'react';
```
ajkl2533, line 1: Content of this file should live in `src/components/ScoreDelta/ScoreDelta.tsx` we are using `index.ts` only for export

ajkl2533 (approved):

PR #778, opened 2023-06-01
feat(ErrorBoundary): add component

julesbarr:


src/components/ErrorBoundary/ErrorBoundary.tsx
```diff
@@ -0,0 +1,123 @@
...
+    if (this?.state.hasError) {
+      return <Component {...this.props} />;
+    } else {
+      return this.props.children;
+    }
+  }
+}
+/* eslint-enable */
+
+export default ErrorBoundary;
```
ajkl2533, line 123: @julesbarr This should have been simple presentational component that will be used by the error boundary component that's already implemented in the platform. We don't have the ambition to implement error boundary functionality in DS

src/components/ErrorBoundary/ErrorBoundary.stories.tsx
```diff
@@ -0,0 +1,87 @@
...
+    </ErrorBoundary>
+  );
+};
+Playground.args = { size: 'md' };
+Playground.parameters = {
+  screenshot: { skip: true },
+};
+
+export const Sizes = ErrorBoundaryTemplate.bind({});
+Sizes.args = { hasButton: true };
```
ajkl2533, line 87: `hasButton` property does not exist

ajkl2533 (changes requested):
Let's get rid of the error boundary implementation part and convert it to the presentational component

ajkl2533 (approved):

PR #780, opened 2023-06-05
fix(Toast): add loading variant

julesbarr:


src/components/_internal/BaseToastBanner/BaseToastBanner.enums.ts
```diff
@@ -3,4 +3,5 @@ export const BaseToastBannerVariants = {
   warn: 'warn',
   error: 'error',
   success: 'success',
+  loading: 'loading',
```
ajkl2533, line 6: This might be problematic since also Banner reuses this enum. We either have to fix the Banner component the same as you did with ProgressBar or we should come up with some other solution.

ajkl2533 (commented):
Please add Loading variant story in Storybook

ajkl2533 (approved):

PR #782, opened 2023-06-06
feat: upgrade React to v18

ajkl2533:
BREAKING CHANGE: From v2 only React18 will be supported


src/components/Datatable/BatchModule/BatchActions/BatchActions.tsx
```diff
@@ -77,7 +76,7 @@ const BatchActions: FC<BatchActionsProps> = ({ actions }) => {
               <BatchActionButton
                 name={action.name}
                 variant={ButtonVariants.text}
-                onClick={action.onClick}
```
julesbarr, line 80: was this removed on purpose ?

src/components/UserAvatar/UserAvatar.tsx
```diff
@@ -96,15 +95,14 @@ const UserAvatar: FC<UserAvatarProps> = forwardRef<
   ) =>
     any(isNotUndefined, [href, to, onClick]) ? (
       <RootUserAvatar
-        ref={ref}
+        ref={ref as unknown as Ref<HTMLButtonElement>}
```
julesbarr, line 98: why do we need this cast ?

src/components/Datatable/BatchModule/BatchActions/BatchActions.tsx
ajkl2533: Screenshot 2023-06-06 at 17 55 37

src/components/UserAvatar/UserAvatar.tsx
ajkl2533: Screenshot 2023-06-06 at 17 57 17

PrincAm (approved):

src/components/Datatable/BatchModule/BatchActions/BatchActions.tsx
ajkl2533: I reworked the component so we can leave the `onClick` there 👍

src/components/UserAvatar/UserAvatar.tsx
ajkl2533: I dropped the mid-conversion to `unknown`, not sure why I added it there. The casting needs to stay there because we need to add the correct type based on if the avatar is interactive or not

Mnau-mnau (approved):

julesbarr (approved):

PR #788, opened 2023-06-28
fix(Text): change default size

julesbarr:


.storybook/image-snapshots/expected/components_forms_Range_Default.png
MateCziner, line 1: looks like the min and max labels became medium, but not the value label of the handle. Previously they were all the same size (Large) so it may make sense to make the handle medium as well

.storybook/image-snapshots/expected/components_typography_Paragraph_Text In Paragraph.png
MateCziner, line 1: 2 apparent font-sizes in the same paragraph, the secondary text seems to be smaller than the rest

.storybook/image-snapshots/expected/components_Accordion_Custom Title Element.png
MateCziner, line 1: in this case only the Company name become smaller, but the body text retained the large size. It should be the other way around, or both should be medium

.storybook/image-snapshots/expected/components_Stepper_Vertical With Expanded Steps.png
MateCziner, line 1: the "done step content" placeholder seem to have retained large size, probably not a real issue though

.storybook/image-snapshots/expected/components_forms_Range_Custom.png
MateCziner, line 1: min max labels became medium, but the handle remained large

.storybook/image-snapshots/expected/components_typography_Text_Shorthand Components.png
MateCziner, line 1: the strong component remained of large size, but the strong tag should not have impact on the font-size, so I guess this should be medium as well

ajkl2533 (approved):
LGTM

PR #814, opened 2023-08-30
fix(ColumnsControl): fix incorrect syntax on PropTypes definition

mrthebunny:
Incorrect PropTypes syntax caused the following warning to show up in console

```
    Warning: Invalid arguments supplied to oneOf, expected an array, got 2 arguments. A common mistake is to write oneOf(x, y, z) instead of oneOf([x, y, z]).

      at printWarning (node_modules/prop-types/factoryWithTypeCheckers.js:23:15)
      at Object.onApply [as oneOf] (node_modules/prop-types/factoryWithTypeCheckers.js:314:11)
      at Object.<anonymous> (node_modules/@securityscorecard/design-system/src/components/Datatable/components/ColumnsControls/ColumnsControls.tsx:110:3)
```


ajkl2533 (approved):
LGTM

PR #821, opened 2023-09-21
feat(StatusPill): add component

Mnau-mnau:


src/components/StatusPill/StatusPill.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+const StyledPillWrapper = Styled(Padbox)`
+  display: inline-block;
+  height: ${pxToRem(24)};
+  min-width: 0;
+  border-radius: ${getRadii('round')};
+  border: 1px solid ${getColor('neutral.200')};
+  background-color: ${getColor('neutral.0')};
+  padding-right: ${pxToRem(6)};
+  padding-top: ${pxToRem(3)};
```
ajkl2533, line 21: I understand why the right padding is different, but why the top is set to 3?

Mnau-mnau: padding xs squish make the component 22 px, square 26, fix size is ok but content is 1 px up

ajkl2533 (approved):

PR #846, opened 2023-10-18
feat(DatatableV2): implement rows per page select

ajkl2533:
Closes UXD-1268

### DIFF
--- a/src/components/DatatableV2/toolbar/Pagination.tsx
+++ b/src/components/DatatableV2/toolbar/Pagination.tsx
@@ -2,61 +2,88 @@ import React from 'react';
 
 import { DatatableInstance } from '../Datatable.types';
 
-export function useDebounce(value, delay) {
-  const [debouncedValue, setDebouncedValue] = React.useState(value);
-
-  React.useEffect(() => {
-    const handler = setTimeout(() => {
-      setDebouncedValue(value);
-    }, delay);
-
-    return () => {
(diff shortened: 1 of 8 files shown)


src/components/DatatableV2/toolbar/Pagination.tsx
```diff
@@ -2,61 +2,88 @@ import React from 'react';
...
+  } = table;
+  const { pagination } = getState();
+  const { pageIndex, pageSize } = pagination;
+
+  const currentPage = pageIndex + 1;
+  const lastPage = getPageCount() - 1;
+  const totalRowCount =
+    options.rowsCount ?? getPrePaginationRowModel().rows.length;
+  const firstRowIndex = pageIndex * pageSize;
+  const lastRowIndex = Math.min(pageIndex * pageSize + pageSize, totalRowCount);
```
Mnau-mnau, line 26: shouldn't this be ... totalRowCount -1?

Mnau-mnau (approved):

src/components/DatatableV2/toolbar/Pagination.tsx
ajkl2533: Nope this is used to display rows range on current page. E.g. I'm on page 2 and I display 20 rows per page so the firstRowIndex will be `21` and the lastRowIndex should be `40`

PR #853, opened 2023-10-23
fix: improve typing

julesbarr:


src/utils/helpers.ts
```diff
@@ -38,13 +38,22 @@ export type ThemeType = {
   padding?: SpacingSizeValue;
 };
 
+type HelperFn<Input, ReturnType> = {
```
ajkl2533, line 41: You can assing default value (`string`) to the second generic param and then you can omit it on other places. It should be relatively safe as we are returning only strings anyway.
```suggestion
type HelperFn<Input, ReturnType = string> = {
```

src/utils/helpers.ts
```diff
@@ -69,69 +74,71 @@ export const getColor: GetColor = curry(
...
+export const getFontWeight: HelperFn<keyof FontWeight, string> = curry(
   (weight: keyof FontWeight, { theme }: ThemeType): string =>
     path(['typography', 'weight', weight], theme),
 );
 
 // getFontSize :: Size -> Props -> string
 // Size - any key of 'size' (src/theme/typography.ts)
 // Props - styled-components props object
-export const getFontSize = curry(
+export const getFontSize: HelperFn<FontSize | string, string> = curry(
```
ajkl2533, line 93: Why are you using the union type here (`FontSize | string`)?  I don't think we allow passing any custom value here. Also if we would it's needed to write it as `... | (string & {})` to make suggestions work.

ajkl2533:
```suggestion
export const getFontSize: HelperFn<keyof FontSize, string> = curry(
```

ajkl2533 (approved):

PR #875, opened 2023-11-29
feat(Checkbox): make checkboxId required

Mnau-mnau:


src/components/forms/Select/Select.styles.tsx
```diff
@@ -485,6 +485,7 @@ export const Option: ComponentType<OptionProps<OptionType, boolean>> = (
       <components.Option {...props}>
         <Inline gap={SpaceSizes.sm}>
           <Checkbox
+            checkboxId="select"
```
ajkl2533, line 488: @Mnau-mnau I guess this will be problem, the checkboxId has to be unique so probably something like 
```
checkboxId={`select-${label}`}

PR #892, opened 2023-12-15
fix(Tooltip): set correct z-index in floating elements

ajkl2533:
Closes UXD-1318

### DIFF
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -8,8 +8,12 @@ import { isFalsy } from 'ramda-adjunct';
 import { Padbox } from '../layout';
 import { getColor, getRadii, pxToRem } from '../../utils';
 import { TooltipProps } from './Tooltip.types';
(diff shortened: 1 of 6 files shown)


src/components/Tooltip/Tooltip.tsx
```diff
@@ -19,7 +23,7 @@ const Content = styled(RadixTooltip.Content)<{ $width: TooltipProps['width'] }>`
   animation-timing-function: cubic-bezier(0.16, 1, 0.3, 1);
   will-change: transform, opacity;
   width: ${pipe(prop('$width'), pxToRem)};
-  z-index: 1;
+  z-index: ${({ $isInFloatingElement }) => ($isInFloatingElement ? 1300 : 1)};
```
julesbarr, line 26: shouldn't we use the `getDepth` helper method?

Mnau-mnau (approved):

src/components/Tooltip/Tooltip.tsx
ajkl2533: yeah we probably should

PR #916, opened 2024-02-15
fix: expose theme colors as css variables

julesbarr:


src/utils/helpers.ts
```diff
@@ -156,3 +156,16 @@ export const abbreviateNumber = (value: number): string =>
   numeral(value).format('0.[00]a').toUpperCase();
 
 export const getShadow = () => 'box-shadow: 0px 4px 12px rgba(0, 0, 0, 0.07);';
+
+export const flattenColors = (colors) => {
```
ajkl2533, line 160: @julesbarr looks good. However, I want to move this function into the `GlobalStyles.tsx` file. Having it exported as another helper function doesn't make much sense.

I would also rename it to `generateColorCSSVars` this function should accept the whole `props` object so you can call it in `:root` as simply as `${generateColorCSSVars}. Also, it would be nice to hide the `map` and `join` logic in the `generateColorCSSVars`.

Also please add unit test.

And one question how does the function handle objects that are deeply nested (more than 2 levels) like 
`{a: {b: {c: {d: 3} } } }`

ajkl2533 (approved):

PR #925, opened 2024-02-29
fix(Signal): update signal component

julesbarr:


src/components/Signal/Signal.tsx
```diff
@@ -85,21 +44,89 @@ const Signal: React.FC<SignalProps> = ({
       {...props}
```
ajkl2533, line 44:
```suggestion
      style={{ overflow: 'visible', ...props.style}}
      {...props}
```
Overflow will prevent squishing icon if the container runs out of space

src/components/Signal/Signal.tsx
```diff
@@ -85,21 +44,89 @@ const Signal: React.FC<SignalProps> = ({
...
-      <g>
-        {paths.map((path, key) => (
-          <path
-            // eslint-disable-next-line react/no-array-index-key
-            key={`${kind}-icon-${key}`}
-            d={path}
-            fill={colors.neutral[0]}
-          />
-        ))}
+      <g style={{ mixBlendMode: 'multiply' }}>
```
ajkl2533, line 47: The `g` tag can be removed from SVG

src/components/Signal/Signal.tsx
```diff
@@ -85,21 +44,89 @@ const Signal: React.FC<SignalProps> = ({
...
+        <path
+          d="M3.5 8C4.3125 8 5 8.6875 5 9.5V14.5C5 15.3438 4.3125 16 3.5 16C2.65625 16 2 15.3438 2 14.5V9.5C2 8.6875 2.65625 8 3.5 8Z"
```
ajkl2533, line 49: Please update bar paths according to Figma. They weren't centred in the viewbox and Mate fixed that.

src/components/Signal/Signal.tsx
```diff
@@ -85,21 +44,89 @@ const Signal: React.FC<SignalProps> = ({
...
+      {...props}
+    />
+  ),
+  [SignalKinds.critical]: (props) => <SeverityIcon {...props} />,
+  [SignalKinds.positive]: ({ size, ...props }: BaseSvgProps) => (
+    <Icon
+      color={ColorTypes.severityPositive}
+      data-testid="ds-severity-icon"
+      name="check-circle-solid"
+      style={{ fontSize: size }}
```
ajkl2533, line 102:
```suggestion
      aria-hidden={false}
      style={{ fontSize: size, verticalAlign: 0 }}
```
We want to add `aria-hidden` prop as we don't want those severity icons to be hidden from screen readers.
The `verticalAlign: 0` reset styling from FA and fixes the misalignment between old and new icons

PR #942, opened 2024-03-26
feat(Button): add iconStart and IconEnd properties

Mnau-mnau:


src/components/Breadcrumbs/BreadcrumbItem.tsx
```diff
@@ -45,7 +45,7 @@ const BreadcrumbItem: React.FC<BreadcrumbItemProps> = ({
         <BreadcrumbLink
           color="secondary"
           href={href}
-          iconName={iconName}
+          iconStart={{ iconName }}
```
ajkl2533, line 48: Please fix the `hasIcon` condition

src/components/_internal/BaseButton/BaseButton.types.ts
```diff
@@ -16,6 +16,15 @@ export type Variants =
...
 
+export interface IconStart {
+  iconName: SSCIcons | string;
+  iconType?: IconTypes | string;
+}
+export interface IconEnd {
+  iconName: SSCIcons | string;
+  iconType?: IconTypes | string;
+}
+
```
ajkl2533, line 27: Merge these two interfaces as:

```ts
export interface BaseIcon {
  // eslint-disable-next-line @typescript-eslint/ban-types
  name: SSCIcons | (string & {});
  type?: Types;
}

File: src/components/_internal/BaseButton/BaseButton.types.ts [hunk 3]
```diff
@@ -29,9 +38,11 @@ export interface BaseButtonProps
   href?: string;
   to?: To;
   className?: string;
-  iconName?: SSCIcons | string;
+  iconStart?: IconStart;
+  iconEnd?: IconEnd;
+  iconName?: SSCIcons | string; // @deprecated Use iconStart property instead
+  iconType?: IconTypes | string; // @deprecated Use iconStart property instead
```
ajkl2533, line 44: I think this won't work as JSDoc comment, it should be like this:
```suggestion
  /** @deprecated Use iconStart property instead */ 
  iconName?: SSCIcons | string;
  /** @deprecated Use iconStart property instead */
  iconType?: IconTypes | string;
```

src/components/Breadcrumbs/BreadcrumbItem.tsx
ajkl2533: No need to change public API here. The issue was that the Breadcrumb link with icon contained also the label, which was regression because label should be hidden in this case

.storybook/image-snapshots/expected/components_Breadcrumbs_With Dropdown.png
ajkl2533, line 1: @Mnau-mnau See here, the issue with the icon on breadcrumb is visible here. the first item should have only icon visible, but after changes you can see also the `Root` label

src/components/Breadcrumbs/BreadcrumbItem.tsx
Mnau-mnau: 🤦‍♀️  I get it

ajkl2533 (approved):

PR #959, opened 2024-04-09
feat(PPT-1969): Implementing title attribute

samuelfajreldines:


ajkl2533 (changes requested):
@samuelfajreldines Code looks good to me, but two things:
1) rebase on top of the `next` branch which is used to prepare releases
2) change the commit message to `feat(Pill): expose "title" property` to follow contribution guidelines

PR #1008, opened 2024-06-21
fix(DatatableV2): use checkbox for general pinning control

silvioprog:


ajkl2533 (approved):
Nice job 👍 
Please update the commit message to `fix(DatatableV2): use checkbox for general pinning control`.

PR #1012, opened 2024-06-26
feat(dialogs): implement focus trap for Modal and Drawer

silvioprog:
## Demos

https://github.com/securityscorecard/design-system/assets/1456829/d190420d-b1ba-4532-92f3-9e9acec8caf5


src/components/Drawer/Drawer.tsx
```diff
@@ -206,10 +207,26 @@ const Drawer = forwardRef<HTMLDivElement, DrawerProps>(
         <Portal>
           {hasBackdrop ? (
             <Overlay placement="right">
-              <DrawerBox {...drawerProps} />
+              <FocusTrap
```
ajkl2533, line 210: I wonder, since the same configuration of the FocusTrap component is used on all places. If it make sense to extract it and put into `_internal` directory so we don't need to repeat the same code everywhere

silvioprog: Fixed

ajkl2533 (approved):

PR #1073, opened 2024-09-02
chore(BaseDateRangePicker): remove prop-types

silvioprog:
Closes UXD-1575.

### DIFF
--- a/src/components/_internal/BaseDateRangePicker/BaseDateRangePicker.tsx
+++ b/src/components/_internal/BaseDateRangePicker/BaseDateRangePicker.tsx
@@ -1,15 +1,9 @@
-import React from 'react';
-import PropTypes from 'prop-types';
 import styled from 'styled-components';
 import { any, prop } from 'ramda';
 import { isNotNull } from 'ramda-adjunct';
 
 import { dateRangePickerStyles } from './styles';
-import {
-  BaseDateRangePickerPropTypes,
-  BaseDateRangePickerProps,
-  BaseDateRangePlaceholderPropTypes,
-} from './BaseDateRangePicker.types';
+import { BaseDateRangePickerProps } from './BaseDateRangePicker.types';
 import { BaseSingleDatePicker } from '../BaseSingleDatePicker';
 import { Inline } from '../../layout/Inline';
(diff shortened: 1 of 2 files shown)


src/components/_internal/BaseDateRangePicker/BaseDateRangePicker.tsx
```diff
@@ -1,15 +1,9 @@
-import React from 'react';
-import PropTypes from 'prop-types';
 import styled from 'styled-components';
 import { any, prop } from 'ramda';
 import { isNotNull } from 'ramda-adjunct';
 
 import { dateRangePickerStyles } from './styles';
-import {
-  BaseDateRangePickerPropTypes,
```
ajkl2533, line 9: You should also remove the custom proptypes definitions that are imported into file (unless it is used also elsewhere). We don't want to end up with unused code.

silvioprog: Good catch. Fixed!

silvioprog: I'm gonna fix the errors.

src/components/Filters/FilterRow/FilterRow.tsx
```diff
@@ -352,7 +351,6 @@ FilterRow.propTypes = {
     PropTypes.string,
     PropTypes.arrayOf(PropTypes.string),
     PropTypes.instanceOf(Date),
-    BaseDateRangePickerPropTypes,
```
ajkl2533, line 355: You shouldn't remove the prop type from here, because you will create a bunch of console errors. That's why I wrote that the custom proptypes should be removed only unless they are not used elsewhere.
In this case you have two options:
1. leave the `BaseDateRangePickerPropTypes` in place and remove it when you remove proptypes from FilterRow, FilterStatePropType and FieldPropTypes
2. remove proptypes where the `BaseDateRangePickerPropTypes` is used

silvioprog: Rebased and fixed! 🙌

ajkl2533 (approved):

PR #1128, opened 2024-09-18
fix(Stack): allow polymorphic behaviour for extended components

ajkl2533:
Previous change introduced accidental breaking change extensively described in https://github.com/styled-components/styled-components/issues/2129. We removed wrapping into React component to revert to previous behaviour.

### DIFF
--- a/src/components/layout/Stack/Stack.tsx
+++ b/src/components/layout/Stack/Stack.tsx
@@ -1,9 +1,4 @@
-import {
-  type ComponentPropsWithRef,
-  type ElementType,
-  type ReactNode,
-  forwardRef,
-} from 'react';
+import { type ReactNode } from 'react';
 import styled from 'styled-components';
 import { prop } from 'ramda';
 import { isNotUndefined } from 'ramda-adjunct';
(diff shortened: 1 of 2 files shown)


src/components/layout/Stack/Stack.tsx
```diff
@@ -28,89 +23,64 @@ export interface StackProps {
...
+  })
+  .attrs<StackProps, StackProps>(
+    ({ gap, justify, isRecursive, className, ...rest }) => ({
+      gap: gap ?? 'none',
+      justify: justify ?? 'stretch',
+      isRecursive: isRecursive ?? false,
+      className: cls(CLX_LAYOUT, className),
+      ...rest,
+    }),
+  )`
```
ajkl2533, line 51: @silvioprog This way we can provide default props directly to styled-component

src/components/layout/Stack/Stack.tsx
```diff
@@ -28,89 +23,64 @@ export interface StackProps {
...
-};
 
-const StackRoot = styled.div<StackRootProps>`
+const Stack = styled.div
+  .withConfig<StackProps>({
+    shouldForwardProp: (propName) =>
+      !['gap', 'justify', 'align', 'splitAt', 'isRecursive', 'theme'].includes(
+        propName,
+      ),
+  })
```
ajkl2533, line 42: @silvioprog It's important to filter our any property used for styling (used inside styled-component definition) to prevent leaking those properties into DOM. The `theme` property is used internally and should be always filtered out.

silvioprog (approved):

PR #1155, opened 2024-09-30
refactor(Text): replace defaultProps property

silvioprog:
Closes UXD-1640.

### DIFF
--- a/src/components/Text/Text.tsx
+++ b/src/components/Text/Text.tsx
@@ -104,12 +104,13 @@ const variants = {
   [TextVariants.context]: contextVariant,
 };
 
-const Text = styled.span.attrs<{ variant: keyof typeof TextVariants }>(
-  (props) => ({
-    ...props,
-    className: cls(CLX_TYPOGRAPHY, props?.className),
-  }),
-)<TextProps>`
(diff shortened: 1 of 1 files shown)


src/components/Text/Text.tsx
```diff
@@ -104,35 +104,27 @@ const variants = {
...
-
 export default Text;
 
 export const Strong = styled(Text).attrs<StrongProps>(() => ({
   as: 'strong',
   isBold: true,
-}))``;
-
-Strong.defaultProps = {
   size: TextSizes.inherit,
```
ajkl2533, line 126:
```suggestion
  size: props.size ?? TextSizes.inherit,
```

silvioprog: What should I put instead?

silvioprog: What would you recommend I use instead?

silvioprog: image

ajkl2533: oh you are right I missed that size is omitted for the Strong

ajkl2533 (approved):

PR #1223, opened 2025-01-21
fix(SelectionToolbar): fix toolbar visibility for manual Pagination UXD-1703

ajkl2533:
### Issue Description
A bug was discovered in the DataTableV2 component related to row selection behavior when using manual (server-side) pagination. The issue manifests in two ways:

- Selection toolbar visibility was incorrectly handled due to a condition error
- Selected row data was being lost when navigating between pages in manually paginated tables

The core problem stems from the `renderRowSelectionActions` property, which expects to receive complete row data objects for selected rows. However, this is impossible to maintain with manual pagination since only the current page's data is available at any given time.

### Solution
Instead of implementing a breaking fix, we decided to formalize the current behavior as a feature with the following changes:

#### New Feature: Row Selection Modes
Added a new property `rowSelectionMode` to DataTable v2 with two options:

- `single-page` (default): Maintains current behavior
- `multi-page`: Enables cross-page selection with ID-based tracking

Closes UXD-1703


Douglasgomes027:
Looks good to me! Clever approach using the Row Selection Modes. 
image

Douglasgomes027 (approved):

PR #1280, opened 2025-03-24
feat(DataTableV2): virtual select all. UXD-1676

mrthebunny:
DataTableV2: Added "virtual select all"

Allows the user to "select" all rows, even if those are not all present in the datatable (ie server-side pagination). 

https://github.com/user-attachments/assets/bead39fd-d68b-471c-83ac-92451b89971b


src/components/DatatableV2/tests/selection.test.tsx
```diff
@@ -392,4 +392,38 @@ describe('DatatableV2/selection', () => {
       expect(screen.getAllByLabelText('Toggle select row')[1]).toBeChecked();
     });
   });
+
+  describe('with "selectAllMode=virtual"', () => {
+    it('should display the "select all" prompt when all rows are selected', async () => {
```
ajkl2533, line 397: @mrthebunny Could you please add case when user selects all row and then deselect one row

ajkl2533:
@mrthebunny Please update the commit message to: `feat(DatatableV2): add virtual select all mode`

julesbarr (approved):

PR #1304, opened 2025-06-02
fix(Callout): add aria attributes. PLAT-529

nasterboost:
## Description

- Ensure the icon is hidden from screen readers.
- Add the aria label prop for the global Callout component.


src/components/Callout/Callout.tsx
```diff
@@ -27,9 +29,11 @@ const Callout = ({
   children,
   icon = 'lightbulb',
   color = 'info',
+  ariaLabel,
```
ev-ssc, line 32: Not an error nor best practice but you could maintain this attribute as optional, just in case

PR #1309, opened 2025-06-12
fix(Breadcrumbs): console error by incorrect key placement. NO_TICKET

mrthebunny:
When a `SeparatorListItem` was added to wrap the separator, the `key` was not moved to the wrapper. Hence it was not being caught by the enclosing the `map` function, causing an `console.error` entry at every render.

This fix moves the `key` to the correct location.

image


Douglasgomes027 (commented):
LGTM

Douglasgomes027 (approved):

PR #1329, opened 2025-09-09
fix(Drawer): dropdown menu click. PLAT-927

nasterboost:
## Description

The Drawer component's focus trap configuration interferes with dropdown menu interactions when dropdown menus are used within the drawer content. This causes the focus trap to deactivate unexpectedly, breaking accessibility and focus management.

### DIFF
--- a/src/components/Drawer/Drawer.tsx
+++ b/src/components/Drawer/Drawer.tsx
@@ -206,7 +206,8 @@ const Drawer = forwardRef<HTMLDivElement, DrawerProps>(
                 const drawer = document.querySelector('[role="dialog"]');
                 return drawer as HTMLElement;
               },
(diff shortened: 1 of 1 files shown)


src/components/Drawer/Drawer.tsx
```diff
@@ -206,7 +206,8 @@ const Drawer = forwardRef<HTMLDivElement, DrawerProps>(
                 const drawer = document.querySelector('[role="dialog"]');
                 return drawer as HTMLElement;
               },
-              clickOutsideDeactivates: hasBackdrop || clickOutsideToHide,
+              clickOutsideDeactivates: true,
```
Douglasgomes027, line 209: image
The `hasBackdrop` prop is being used in many places of ui-core. Is this intentional?

PR #1338, opened 2025-09-22
feat(toggle): add focus support for keyboard. PLAT-947

loengrysequeira-ssc:
## Changes
* Add focus to the toggle item to avoid losing focus when changing the value.

https://github.com/user-attachments/assets/a41f9a70-8ead-43ee-92cd-82d9aca2f98a


Douglasgomes027 (approved):
LGTM

PR #1339, opened 2025-09-24
feat(datatablev2): change scroll button position. PLAT-989

loengrysequeira-ssc:
## Changes

Corrected the focus visibility in the horizontal scroll using existing functionality. 
Updated outline styles to be more prominent.
Fullscreen Support: Works in both normal and full-screen table modes

### DIFF
--- a/src/components/DatatableV2/table/TableRoot.tsx
+++ b/src/components/DatatableV2/table/TableRoot.tsx
@@ -1,5 +1,41 @@
 import styled from 'styled-components';
 
+export const TableContainer = styled.div`
+  position: relative;
+
+  /* Keyboard accessible horizontal scrollbar */
+  .ds-table-horizontal-scrollbar {
+    position: absolute;
+    bottom: 0;
+    left: 0;
+    right: 0;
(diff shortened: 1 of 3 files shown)


src/components/DatatableV2/table/TableRoot.tsx
```diff
@@ -58,6 +34,36 @@ const TableRoot = styled.div`
...
+  }
+  &::-webkit-scrollbar-thumb {
+    border-radius: var(--sscds-radii-rounded);
+    background-color: var(--sscds-color-primary-9);
+    border: 8px solid var(--sscds-color-neutral-0);
+  }
+  &::-webkit-scrollbar-track-piece {
+    border-radius: var(--sscds-radii-rounded);
+    background-color: var(--sscds-color-neutral-4);
+    border: 8px solid var(--sscds-color-neutral-0);
```
Douglasgomes027, line 65: Shouldn’t we move this to a token? It seems we’re going to need to reuse it

Douglasgomes027 (approved):

